PR #67, opened 2021-04-16
INGK-211-integrate-ingenialogger

oriolbv:


ingenialink/canopen/net.py
```diff
@@ -109,17 +108,16 @@ def __init__(self, device=None, channel=0,
                                        channel=self.__channel,
                                        bitrate=self.__baudrate)
             except Exception as e:
```
GullinBustin, line 110: Exception is too generic.

ingenialink/canopen/net.py
```diff
@@ -109,17 +108,16 @@ def __init__(self, device=None, channel=0,
...
-                print('Exception trying to connect: ', e)
+                logger.error('Failed trying to connect. Exception: %s', str(e))
```
GullinBustin, line 111: Cast error to string is not necessary, you can use it directly.

ArnauP (approved):

ArnauP:
This is already implemented in INGK-222

PR #177, opened 2022-06-20
INGK-338 Adapt python libraries to work linux

MartinALL91:


requirements/dev-requirements.txt
```diff
@@ -0,0 +1,14 @@
+-r requirements.txt
+qtpy==1.9.0
+qtmodern==0.2.0
+qtawesome==1.0.3
+matplotlib==3.3.4
+pycodestyle==2.6.0
```
GullinBustin, line 6: I think we can remove from dev requirements these modules:
- qtpy==1.9.0
- qtmodern==0.2.0
- qtawesome==1.0.3
- matplotlib==3.3.4
- pycodestyle==2.6.0

PR #178, opened 2022-06-21
INGK-461 Avoid importing python can lib in linux

MartinALL91:


ingenialink/__init__.py
```diff
@@ -23,12 +23,15 @@
...
-from .canopen.dictionary import CanopenDictionary
+try:
+    from .canopen.servo import CanopenServo
+    from .canopen.network import CanopenNetwork, CAN_DEVICE, CAN_DEVICE, \
+        CAN_BAUDRATE
+    from .canopen.poller import CanopenPoller
+    from .canopen.register import CanopenRegister
+    from .canopen.dictionary import CanopenDictionary
+except ImportError:
+    print("Error importing python-can library, not supported on Linux.")
```
GullinBustin, line 34: Use logger

ingenialink/__init__.py
```diff
@@ -23,12 +23,15 @@
...
-from .canopen.poller import CanopenPoller
-from .canopen.register import CanopenRegister
-from .canopen.dictionary import CanopenDictionary
+try:
+    from .canopen.servo import CanopenServo
+    from .canopen.network import CanopenNetwork, CAN_DEVICE, CAN_DEVICE, \
+        CAN_BAUDRATE
+    from .canopen.poller import CanopenPoller
+    from .canopen.register import CanopenRegister
+    from .canopen.dictionary import CanopenDictionary
```
GullinBustin, line 32: I'm not sure if I like it. What happens if someone tries to call some of these classes, such as CAN_DEVICE, for example?

MartinALL91: An ImportError will be raised. What is your suggestion to solve this problem?

GullinBustin (approved):

PR #182, opened 2022-06-21
INGK-320 Update docstring

MartinALL91:


ingenialink/ethercat/network.py
```diff
@@ -87,6 +87,8 @@ def load_firmware(self, fw_file, target=1, boot_in_app=True):
         if not os.path.isfile(fw_file):
             raise FileNotFoundError('Could not find {}.'.format(fw_file))
 
+        if boot_in_app is None:
+            boot_in_app = fw_file.endswith(FILE_EXT_SFU)
```
GullinBustin, line 91: In my opinion, if boot_in_app is None and the filename does not end with SFU or LFU, that should raise an exception, because you don't know what file is and user does not give you any info about boot in app

GullinBustin (changes requested):

ingenialink/ethercat/network.py
MartinALL91: Ok, will do.

GullinBustin (approved):

PR #195, opened 2022-07-12
INGK-407 Add checks to drive IP, gateway IP and subnet mask

MartinALL91:


ingenialink/ethernet/servo.py
```diff
@@ -86,7 +89,26 @@ def change_tcp_ip_parameters(self, ip_address, subnet_mask, gateway):
...
+            NetmaskValueError: If the subnet_mask is not a valid
+            netmask.
+
         """
+        try:
+            drive_ip = ipaddress.ip_address(ip_address)
+            gateway_ip = ipaddress.ip_address(gateway)
+            net = ipaddress.IPv4Network(f'{drive_ip}/{subnet_mask}', strict=False)
+        except (ValueError, NetmaskValueError) as e:
+            raise e
```
GullinBustin, line 106: You are catching the exception and raising it. Simply remove the try and catch and you will obtain the same behavior.

GullinBustin: It could be done if you want to log it or similar, like this example (but maybe it is not necessary here, what do you think?): https://stackoverflow.com/questions/2052390/manually-raising-throwing-an-exception-in-python

```logger = logging.getLogger(__name__)

try:
    do_something_in_app_that_breaks_easily()
except AppError as error:
    logger.error(error)
    raise                 # just this!
    # raise AppError      # Don't do this, you'll lose the stack trace!```

GullinBustin (changes requested):

ingenialink/ethernet/servo.py
MartinALL91: I did this to explicitly know where the error is coming from. But you're right, if I remove the try/catch we will get the same behavior.

GullinBustin (approved):

PR #209, opened 2022-10-04
INGK-502 Implement read write ethernet servo using python ethernet lib

MartinALL91:
In this PR there are changes from INGK-501 and INGK-502


GullinBustin:
Ensure the old unit test works or create new tests if necessary

GullinBustin (changes requested):

MartinALL91:
> Ensure the old unit test works or create new tests if necessary

Once I'm done with all the Ethernet implementation I will create/adjust the tests.

ingenialink/canopen/servo.py
```diff
@@ -327,7 +327,7 @@ def read(self, reg, subnode=1):
         """
         _reg = self._get_reg(reg, subnode)
         raw_read = self._read_raw(reg, subnode)
-        value = self.__convert_bytes_to_dtype(raw_read, _reg.dtype)
+        value = convert_bytes_to_dtype(raw_read, _reg.dtype)
```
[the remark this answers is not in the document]
MartinALL91: Done

WoodElv (approved):

ingenialink/utils/_utils.py
```diff
@@ -339,3 +341,66 @@ def set_logger_level(level):
...
+            bytes_length = 2
+            signed = True
+        elif dtype == REG_DTYPE.U32:
+            bytes_length = 4
+        elif dtype == REG_DTYPE.S32:
+            bytes_length = 4
+            signed = True
+        data = data.to_bytes(bytes_length,
+                             byteorder='little',
+                             signed=signed)
```
GullinBustin, line 405: Maybe you can write this function more clear, like:
``` python
if dtype == REG_DTYPE.DOMAIN:
    return data
if dtype == REG_DTYPE.FLOAT:
    return data = struct.pack('f', float(data))
__dtype_value = {
    REG_DTYPE.U8: (1, False),
    REG_DTYPE.S8: (1, True)
    REG_DTYPE.U16: (2, False),
    REG_DTYPE.S16: (2, True),
    REG_DTYPE.U32: (4, False),
    REG_DTYPE.S32: (4, True)
}
bytes_length, signed = __dtype_value[dtype]
...
```
Or something like this, what do you think?

GullinBustin: In this case, with str type will raise an exception.

GullinBustin (changes requested):

ingenialink/utils/_utils.py
MartinALL91: I like it, I will change it.

GullinBustin (approved):

PR #222, opened 2022-10-18
INGK-529: cyclic import between  utils and register module

WoodElv:


GullinBustin:
__init__.py file also imports REG_DTYPE, REG_ACCESS, REG_PHY from IPB/register.py

ingenialink/enums/register.py
```diff
@@ -0,0 +1,55 @@
...
+    """Unsigned 32-bit integer."""
+    S32 = 5
+    """Signed 32-bit integer."""
+    U64 = 6
+    """Unsigned 64-bit integer."""
+    S64 = 7
+    """Signed 64-bit integer."""
+    FLOAT = 8
+    """Float."""
+    STR = 9
```
MartinALL91, line 24: In [ingenialink](https://github.com/ingeniamc/ingenialink/blob/master/include/public/ingenialink/registers.h#L39) the value of STR is 10

MartinALL91 (approved):

ingenialink/enums/register.py
WoodElv: It doesn't matter because when a register object is initialized, it depends on how this enum is defined, there is not anything that makes how to define the enum in a mandatory way.

MartinALL91 (approved):

GullinBustin (approved):

MartinALL91 (approved):

PR #225, opened 2022-10-19
INGK-516 Unify ethernet servo and canopen servo

MartinALL91:


ingenialink/canopen/servo.py
```diff
@@ -20,130 +15,6 @@
...
-    identifier='', units='', subnode=0, idx=0x58E4, subidx=0x00, cyclic='CONFIG',
-    dtype=REG_DTYPE.U16, access=REG_ACCESS.RO
-)
-
-MONITORING_ACTUAL_NUMBER_BYTES = CanopenRegister(
-    identifier='', units='', subnode=0, idx=0x58B7, subidx=0x00, cyclic='CONFIG',
-    dtype=REG_DTYPE.U32, access=REG_ACCESS.RO
-)
-
 MONITORING_DATA = CanopenRegister(
```
GullinBustin, line 17: Why you didn't move MONITORING_DATA, MONITORING_DISTURBANCE_VERSION, and DIST_DATA registers inside the class?

ingenialink/servo.py
```diff
@@ -195,123 +188,1050 @@ def stop(self):
...
+        """
+        try:
+            return self.read(register_coco, subnode=0)
+        except ILError:
+            pass
+
+        try:
+            return self.read(register_moco, subnode=1)
+        except ILError:
+            pass
```
GullinBustin, line 1010: If not exist COCO nor MOCO register it returns None, is it the expected behavior?

GullinBustin (changes requested):

ingenialink/canopen/servo.py
MartinALL91: Because it was not necessary. For coherence, I will move all the registers within the class.

ingenialink/servo.py
MartinALL91: This cannot happen as the register_coco and register_moco registers are defined in the class.

GullinBustin: But it can file with a timeout or similar, in this case the error will not be raise and a None will be return

MartinALL91: Changed

WoodElv (changes requested):

WoodElv (approved):

GullinBustin (approved):

ingenialink/servo.py
```diff
@@ -195,123 +188,1052 @@ def stop(self):
...
+            all the parameters.
+
+        Raises:
+            ILError: Invalid subnode.
+            ILObjectNotExist: Failed to write to the registers.
+
+        """
+        if subnode is None:
+            # Restore all
+            self.write(reg=self.RESTORE_COCO_ALL,
```
GullinBustin, line 378: Function write and property RESTORE_COCO_ALL don't exist in Servo class. Functions should be added as abstract functions and properties as abstract properties (https://stackoverflow.com/questions/5960337/how-to-create-abstract-properties-in-python-abstract-classes) or defined as None in the parent (abstract properties seems overkill for me, what do you think?)

GullinBustin (changes requested):

ingenialink/servo.py
MartinALL91: Done

GullinBustin (approved):

PR #226, opened 2022-10-20
INGK-526 delete ffi and lib dependencies for servos modules

WoodElv:


ingenialink/servo.py
```diff
@@ -600,12 +599,13 @@ def status_word_wait_change(self, status_word, timeout, subnode=1):
         start_time = int(round(time.time() * 1000))
         actual_status_word = self.read(self.STATUS_WORD_REGISTERS[subnode],
                                        subnode=subnode)
+        # Operation time out
+        il_etimedout = -3
```
MartinALL91, line 603: I think this value should the defined some place else

GullinBustin (changes requested):

GullinBustin (approved):

MartinALL91 (approved):

PR #230, opened 2022-11-04
INGK-540 Add multi slave support

MartinALL91:
The EoE service must be started manually after the drives have been configured. This is because the feature to stop the service is not yet implemented.

### DIFF
--- a/ingenialink/eoe/network.py
+++ b/ingenialink/eoe/network.py
@@ -8,24 +8,24 @@ from ingenialink.exceptions import ILTimeoutError, ILIOError, ILError
 
 
 class EoECommand(Enum):
-    INIT = '0'
-    SCAN = '1'
-    CONFIG = '2'
-    START = '3'
-    STOP = '4'
+    INIT = "0"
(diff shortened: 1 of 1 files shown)


WoodElv (approved):

ingenialink/eoe/network.py
```diff
@@ -144,29 +160,29 @@ def _initialize_eoe_service(self):
...
                                           ip_bytes)
         try:
             self._send_command(msg)
         except (ILIOError, ILTimeoutError) as e:
-            raise ILError(f"Failed to configure slave {node} with IP "
+            raise ILError(f"Failed to configure slave {slave_id} with IP "
                           f"{ip_address}.") from e
 
-    def _start_eoe_service(self):
+    def start_eoe_service(self):
```
GullinBustin, line 185: Maybe you can implement a message to ask if the service was started.

GullinBustin (changes requested):

GullinBustin (approved):

PR #231, opened 2022-11-08
INGK-541 Implement stop eoe service feature

MartinALL91:
There is no need to stop the service. 

The solution is to just perform the following operations the first time the service is started:

- Initialize the network interfaces
-  Create the virtual adapter
-  Initialize the SOEM library
-  Start the forwarder
-  Start the EtherCAT master

Then, each time the service is started again only perform the configuration of the slaves.

### DIFF
--- a/ingenialink/eoe/network.py
+++ b/ingenialink/eoe/network.py
@@ -54,15 +54,7 @@ class EoENetwork(EthernetNetwork):
         Returns:
             EthernetServo: Instance of the servo connected.
 
-        Raises:
-            ILError: If a connection is attempted on a slave with an ID
-            greater than the number of connected slaves.
-
         """
-        _num_conn_slaves = self.scan_slaves()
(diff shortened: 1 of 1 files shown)


WoodElv (approved):

ingenialink/eoe/network.py
```diff
@@ -56,14 +56,15 @@ def connect_to_slave(self, slave_id, ip_address, dictionary=None,
 
         """
         self._configure_slave(slave_id, ip_address)
+        if not self._eoe_service_started:
+            self._start_eoe_service()
```
GullinBustin, line 60: What happens if you run a script twice and the service was started in the first run?

MartinALL91: Nothing, you can re-run the script without a problem.

GullinBustin (approved):

PR #232, opened 2022-11-09
INGK-520 Update save_configuration function

pzinemanas:
This PR updates the `save_configuration` function so the xcf file is created based on the registers stored at the dictionary. It also improves the test related to this function so not only checks if the file exists.

### DIFF
--- a/tests/servo/test_servo.py
+++ b/tests/servo/test_servo.py
@@ -49,11 +49,12 @@ def test_save_configuration(connect_to_slave):
         registers = servo.dictionary.registers(subnode=subnode)
 
         assert reg_id in registers
-        address = int(saved_register.attrib.get('address'), 16)
-        assert registers[reg_id].address == address
 
         storage = saved_register.attrib.get('storage')
-        storage == registers[reg_id].storage
+        if storage is not None:
+            assert storage == str(registers[reg_id].storage)
+        else:
+            assert registers[reg_id].storage is None
 
         access = saved_register.attrib.get('access')
         assert registers[reg_id].access == servo.dictionary.access_xdf_options[access]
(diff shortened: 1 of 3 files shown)


tests/servo/test_servo.py
```diff
@@ -1,18 +1,68 @@
...
+        if subnode not in subnodes:
+            continue
+
+        reg_id = saved_register.attrib.get('id')
+        registers = servo.dictionary.registers(subnode=subnode)
+
+        assert reg_id in registers
+
+        storage = saved_register.attrib.get('storage')
+        if storage is not None:
```
MartinALL91, line 54: With Python 3.8+ you could do:

```
reg_storage = registers[reg_id].storage
if (saved_storage := saved_register.attrib.get('storage')) is not None:
    reg_storage = str(reg_storage)
assert reg_storage == saved_storage

MartinALL91 (approved):

tests/servo/test_servo.py
pzinemanas: Interesting, but the library supports py36 and 38 as well, right?

MartinALL91: Yes, the comment was just to show how we could simplify this part of the code by using the walrus operator.

GullinBustin:
As you say in the comment, remove  `servo.__update_single_axis_dict` and `servo.__update_multiaxis_dict` functions

GullinBustin (changes requested):
As you say in the comment, remove `servo.__update_single_axis_dict` and `servo.__update_multiaxis_dict` functions

GullinBustin (approved):

PR #233, opened 2022-11-09
INGK-543 - Remove registers with address_type=NVM_NONE or access!=rw from the xcf file

pzinemanas:
This PR:

- Defines a new attribute `Register.address_type`
- Creates a `REG_ADDRESS_TYPE `enum to define the list of possible `address_type`.
- Reads the `address_type `from the xdf file.
- When saving the xcf file, avoids the registers with `address_type=NVM_NONE` and `access != rw`
- Updates tests accordingly

### DIFF
--- a/ingenialink/dictionary.py
+++ b/ingenialink/dictionary.py
@@ -279,9 +279,9 @@ class Dictionary(ABC):
 
         Raises:
             KeyError: If the register doesn't have an identifier.
-            ILValueError: If the register data type is invalid.
-            ILAccessError: If the register access type is invalid.
-            ILUnsupportedRegisterValue: If the register address type is invalid.
+            ValueError: If the register data type is invalid.
+            ValueError: If the register access type is invalid.
+            ValueError: If the register address type is invalid.
             KeyError: If some attribute is missing.
 
         """
@@ -309,8 +309,8 @@ class Dictionary(ABC):
             if dtype_aux in self.dtype_xdf_options:
                 current_read_register[self.AttrRegDict.DTYPE] = self.dtype_xdf_options[dtype_aux]
             else:
-                raise exc.ILValueError(f'The data type {dtype_aux} does not exist for the register: '
-                                       f'{current_read_register["identifier"]}')
+                raise ValueError(f'The data type {dtype_aux} does not exist for the register: '
+                                 f'{current_read_register["identifier"]}')
 
             # Access type
             access_aux = register.attrib['access']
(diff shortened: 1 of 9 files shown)


ingenialink/dictionary.py
```diff
@@ -305,6 +314,15 @@ def _read_xdf_register(self, register):
                 raise exc.ILAccessError(f'The access type {access_aux} does not exist for the register: '
                                         f'{current_read_register[self.AttrRegDict.IDENTIFIER]}')
 
+            # Address type
+            address_type_aux = register.attrib['address_type']
+
+            if address_type_aux in self.address_type_xdf_options:
+                current_read_register[self.AttrRegDict.ADDRESS_TYPE] = self.address_type_xdf_options[address_type_aux]
+            else:
+                raise exc.ILAccessError(f'The address type {address_type_aux} does not exist for the register: '
```
MartinALL91, line 323: [This ](https://github.com/ingeniamc/ingenialink-python/blob/master/ingenialink/exceptions.py#L132)exception looks better suited for this case. What do you think?

ingenialink/servo.py
```diff
@@ -381,6 +381,8 @@ def save_configuration(self, config_file, subnode=None):
         for subnode in subnodes:
             registers_dict = self.dictionary.registers(subnode=subnode)
             for reg_id, register in registers_dict.items():
+                if (register.address_type == REG_ADDRESS_TYPE.NVM_NONE) | (register.access != REG_ACCESS.RW):
```
MartinALL91, line 384: Why not just use the normal or operator?

MartinALL91 (changes requested):

ingenialink/dictionary.py
MartinALL91: The exception is not documented.

pzinemanas: > [This ](https://github.com/ingeniamc/ingenialink-python/blob/master/ingenialink/exceptions.py#L132)exception looks better suited for this case. What do you think?

I don't know, it's not a register value. Should we create a new exception for this case?

MartinALL91 (approved):

ingenialink/dictionary.py
```diff
@@ -268,6 +277,13 @@ def _read_xdf_register(self, register):
             dict: The current register which it has been reading
             None: When at least a mandatory attribute is not in a xdf file
 
+        Raises:
+            KeyError: If the register doesn't have an identifier.
+            ILValueError: If the register data type is invalid.
+            ILAccessError: If the register access type is invalid.
+            ILUnsupportedRegisterValue: If the register address type is invalid.
```
GullinBustin, line 284: We use IL exceptions when the exceptions are related to a drive error. Here you can simply use ValueError or another standard python exception.

ingenialink/servo.py
```diff
@@ -381,6 +381,8 @@ def save_configuration(self, config_file, subnode=None):
         for subnode in subnodes:
             registers_dict = self.dictionary.registers(subnode=subnode)
             for reg_id, register in registers_dict.items():
+                if (register.address_type == REG_ADDRESS_TYPE.NVM_NONE) or (register.access != REG_ACCESS.RW):
```
GullinBustin, line 384: Is the second condition duplicated in line 390?

`if access_ops[register.access] == 'rw':`

GullinBustin (changes requested):

GullinBustin (approved):

PR #235, opened 2022-11-15
INGK-547 Define enums_count as len(enums) at the Register

pzinemanas:
Bugs: 

- `enums_count` is never defined when instancing the `Register` class. So, `Register.enums_count=0` even if `enums!=None`. 
- `enum` is a one-element list of dict where the dict is the enum dict. A simple dict should work.

Fix:
- Just define `enums_count=len(enums)` if the user doesn't define the value of `enums_count`
- Convert `enum` type from list of dict to dict.

### DIFF
--- a/ingenialink/register.py
+++ b/ingenialink/register.py
@@ -109,7 +109,6 @@ class Register(ABC):
 
     def __config_enums(self):
         aux_enums = []
-        #for enum in self._enums:
         for key, value in self._enums.items():
             dictionary = {
                 'label': value,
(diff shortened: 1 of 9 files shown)


ingenialink/register.py
```diff
@@ -110,13 +109,13 @@ def __config_range(self, reg_range):
...
     def __config_enums(self):
         aux_enums = []
-        for enum in self._enums:
-            for key, value in enum.items():
-                dictionary = {
-                    'label': value,
-                    'value': int(key)
-                }
-                aux_enums.append(dictionary)
+        #for enum in self._enums:
```
MartinALL91, line 112: Remove commented line

MartinALL91 (approved):

PR #236, opened 2022-11-15
INGK-549 Create pull_request_template.md

pzinemanas:
Proposal for PR template.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -16,8 +16,9 @@ Please add a description and delete options that are not relevant.
 - [ ] Change 2 (description)
 
 
-### Test
-- [ ] Run tests
+### Tests
+- [ ] Add new unit tests if it applies.
+- [ ] Run tests.
 
 Please describe the tests that you ran to verify your changes if it applies. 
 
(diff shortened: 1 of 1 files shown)


WoodElv (approved):

.github/pull_request_template.md
```diff
@@ -0,0 +1,31 @@
+### Pull request name 
+
+While your working on your PR please add the [WIP] indication in the title, to avoid reviewing before it's ready. Also, please try to have a descriptive title.
```
MartinALL91, line 3: In my opinion, a PR should only be opened once the issue is solved. Therefore I think that this field is not necessary. What do you guys think @GullinBustin @WoodElv?

pzinemanas: Interesting. I used to create a PR just when the branch is created so we can start a discussion about the changes, or some collaborators can start reviewing even the PR is not finished. In fact, github has a [draft tag](https://github.blog/2019-02-14-introducing-draft-pull-requests/) that we could use to do this. Also, the checklist at the template illustrates how advanced is the PR.

pzinemanas: Agreed, @GullinBustin @WoodElv any thoughts?

WoodElv: Yes, I think it's a good idea the using of PR in draft state.

GullinBustin (approved):

MartinALL91 (approved):

PR #237, opened 2022-11-16
INGK-548 Avoid using the EDS file for Canopen

pzinemanas:
### Description

It seems the EDS file is never used either by the canopen library nor by ingenialink. The canopen library loads the optional EDS file when [`canopen.Network.add_node`](https://canopen.readthedocs.io/en/latest/network.html?highlight=canopen.Network.add_node#canopen.Network.add_node) is called, but this dictionary is never used. Only [`RemoteNode.load_configuration`](https://canopen.readthedocs.io/en/latest/network.html?highlight=canopen.Network.add_node#canopen.RemoteNode.load_configuration) function would load the dictionary to the drive, but this function is never called.

Fixes # INGK-548

### Type of change

- [x] Remove eds files from the CanopenServo class.
- [x] Remove eds files from examples.

### Tests
- [x] Run tests.
- [x] Test ingeniamotion and motionlab.

We have run the test of ingenialink without using the EDS file and the test have passed. We have also tested some functionalities of ingeniamotion and motionlab and everything seems to be working correctly, but more tests are needed to be sure.

### Documentation


pzinemanas:
This PR would break backward compatibility as it is now. Should we keep backward compatibility in these cases?

WoodElv (approved):

MartinALL91 (approved):

PR #241, opened 2022-11-24
INGK-561 Improve canopen tests

pzinemanas:
### Description

This PR improves the tests of canopen protocol

Fixes # INGK-561

### Type of change

- [x] Add a virtual device for canopen
- [x] Improve the tests of canopen_servo
- [x] Improve tests of canopen_network
- [x] Move some tests from test_canopen_servo to test_servo since those tests were duplicated or worked for all protocols.
- [x] Add a test of load_firmware

### Tests
- [x] Run tests and obtain better coverage

### DIFF
--- a/tests/canopen/test_canopen_network.py
+++ b/tests/canopen/test_canopen_network.py
@@ -101,3 +101,19 @@ def test_setup_and_teardown_connection(virtual_network):
     assert virtual_network._connection is not None
     virtual_network._teardown_connection()
     assert virtual_network._connection is None
+
+
+@pytest.mark.skip
+def test_load_firmware(connect_to_slave, read_config):
+    # TODO: Fix load_firmware method to work independently of status listeners
+    servo, net = connect_to_slave
+    assert servo is not None and net is not None
+    assert len(net.servos) == 1
+    fw_version = servo.read('DRV_ID_SOFTWARE_VERSION')
+    protocol_contents = read_config['canopen']
+
+    net.load_firmware(protocol_contents['node_id'], protocol_contents['test_fw_file'])
+    new_fw_version = servo.read('DRV_ID_SOFTWARE_VERSION')
+    
+    assert new_fw_version != fw_version
+    net.disconnect_from_slave(servo)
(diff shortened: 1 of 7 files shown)


tests/canopen/test_canopen_network.py
```diff
@@ -11,6 +35,33 @@ def test_connect_to_slave(connect_to_slave):
     assert fw_version is not None and fw_version != ''
 
 
+@pytest.mark.canopen
+def test_connect_to_slave_target_not_in_nodes(read_config):
+    protocol_contents = read_config['canopen']
```
MartinALL91, line 40: Maybe we should create a fixture that provides us with the configuration content of the corresponding protocol. In case the protocol name is changed.

tests/canopen/test_canopen_network.py
```diff
@@ -19,3 +70,50 @@ def test_scan_slaves(read_config):
...
+    servo, net = connect_to_slave
+    assert servo is not None and net is not None
+    assert len(net.servos) == 1
+    fw_version = servo.read('DRV_ID_SOFTWARE_VERSION')
+    protocol_contents = read_config['canopen']
+
+    net.load_firmware(protocol_contents['node_id'], protocol_contents['test_fw_file'])
+    new_fw_version = servo.read('DRV_ID_SOFTWARE_VERSION')
+    
+    assert new_fw_version != fw_version
```
MartinALL91, line 118: We have to make sure that the current firmware version is different than the firmware version to be loaded.

MartinALL91: Furthermore, do we want to switch back to the previous firmware after this test?

tests/test_servo.py
```diff
@@ -1,9 +1,13 @@
 import os
 import pytest
+import shutil
+from pathlib import Path
+
```
MartinALL91, line 5: xml is a standard library. This line should be removed.

tests/test_servo.py
```diff
@@ -83,6 +87,64 @@ def test_load_configuration(connect_to_slave, read_config, pytestconfig):
...
+    assert servo is not None and net is not None
+    
+    protocol = pytestconfig.getoption("--protocol")
+    filename = read_config[protocol]['load_config_file']
+    with pytest.raises(ValueError):
+        servo.load_configuration(filename, subnode=subnode)
+
+
+@pytest.mark.canopen
+@pytest.mark.ethernet
```
MartinALL91, line 119: Why not test this with Ethercat?

MartinALL91 (approved):

GullinBustin (approved):

PR #249, opened 2022-11-30
INGK-578 Launch ingenialink tests

MartinALL91:
Update Jenkinsfile and config.json in order to run tests via Jenkins.


Jenkinsfile
```diff
@@ -1,68 +1,140 @@
...
+            exit /b 0
+        '''
+    }
+
+    stage('Run Ethernet tests') {
+        bat '''
+            venv\\Scripts\\python.exe -m pytest tests --protocol ethernet
+            exit /b 0
+        '''
+    }
```
pzinemanas, line 73: We need to run no-connection tests as well:

```
    stage('Run no-connection tests') {
        bat '''
            venv\\Scripts\\python.exe -m pytest tests
            exit /b 0
        '''
    }
````

**Review by pzinemanas — CHANGES_REQUESTED** [review]

File: Jenkinsfile (hunk 1 above)
Comment (MartinALL91) [reply Jenkinsfile:73]: Ok, I'll add it.

File: Jenkinsfile [hunk 2]
```diff
@@ -1,68 +1,195 @@
...
+            }
+            stages {
+                stage('Clone repository') {
+                    steps {
                         bat """
-                            py${version}\\Scripts\\python.exe -m pycodestyle --first ingenialink/ --config=setup.cfg
+                            cd C:\\Users\\ContainerAdministrator
+                            git clone https://github.com/ingeniamc/ingenialink-python.git
+                            cd ingenialink-python
+                            git checkout ${env.GIT_COMMIT}
```
GullinBustin, line 27: I think the repository is already cloned, you only should do a checkout with "checkout scm"

GullinBustin (changes requested):

WoodElv (approved):

Jenkinsfile
MartinALL91: That is the first thing I tried, If we do It like that we get an error when installing the lib.

MartinALL91: Check the output of the Jenkins build #12 for the branch test-INGK-578.

GullinBustin: Ok, we will try to improve it in the future

GullinBustin (approved):

pzinemanas (approved):

PR #250, opened 2022-12-01
INGK-581 Move registers from EthernetServo and CanopenServo to the parent.

pzinemanas:
### Description

This PR moves the registers that are defined in the Servo children to the parent class. Besides, the registers are defined using the ID instead of a Register instance.

Fixes # INGK-581

### Type of change

- [x] Define registers at Servo using the id instead of the Register instance (note that the dictionary is needed to do monitoring and disturbance)
- [x] Define the registers at the Servo parent instead of its children if possible.
- [x] Fix an issue at the `Drive.restore_parameters` method when restoring MOCO parameters.
- [ ] Make dictionary mandatory at Servo.

### Tests


ingenialink/servo.py
```diff
@@ -66,27 +66,27 @@ class Servo:
...
-    MONITORING_NUMBER_MAPPED_REGISTERS = None
-    MONITORING_BYTES_PER_BLOCK = None
-    MONITORING_ACTUAL_NUMBER_BYTES = None
+    STATUS_WORD_REGISTERS = "DRV_STATE_STATUS"
+    RESTORE_COCO_ALL = "DRV_RESTORE_COCO_ALL"
+    RESTORE_MOCO_ALL_REGISTERS = "DRV_RESTORE_MOCO_ALL"
+    STORE_COCO_ALL = "DRV_STORE_COCO_ALL"
+    STORE_MOCO_ALL_REGISTERS = "DRV_STORE_MOCO_ALL"
+    CONTROL_WORD_REGISTERS = "DRV_STATE_CONTROL"
+    SERIAL_NUMBER_REGISTERS = ["DRV_ID_SERIAL_NUMBER_COCO", "DRV_ID_SERIAL_NUMBER"]
```
MartinALL91, line 75: Update docstring of `__read_coco_moco_register`, the args type has changed.

ingenialink/servo.py
```diff
@@ -284,10 +284,10 @@ def restore_parameters(self, subnode=None):
...
-        elif subnode > 0 and subnode in self.RESTORE_MOCO_ALL_REGISTERS:
+        elif subnode > 0:
             # Restore axis
-            self.write(reg=self.RESTORE_COCO_ALL,
-                       data=self.RESTORE_MOCO_ALL_REGISTERS[subnode],
+            self.write(reg=self.RESTORE_MOCO_ALL_REGISTERS,
+                       data=PASSWORD_RESTORE_ALL,
                        subnode=subnode)
             logger.info(f'Restore subnode {subnode} successfully done.')
         else:
```
MartinALL91, line 293: I would add the subnode value in the error message below.

MartinALL91 (approved):

ingenialink/servo.py
pzinemanas: Done!

GullinBustin (approved):

PR #251, opened 2022-12-01
INGK-579 Avoid repeated code between the Servo children

pzinemanas:
### Description

This PR moves some functions of the Servo children to the Servo parent. This simplifies the code and avoids possible bugs when re-defining abstract methods in the children. Only two methods are kept as abstract methods:  `read_raw` and `write_raw`. Those are protocol-dependent methods to write and read byte arrays.

Fixes # INGK-579

### Type of change

- [x] Add a `DICTIONARY_CLASS` constant at the `Servo` class, which stores the `Dictionary` class related to the `Servo` (e.g., `CanopenDictionary` for `CanopenServo`).
- [x]  Add a `MAX_WRITE_SIZE ` constant at the `Servo` class, which stores the disturbance max size for each protocol.
- [x] Creates `read_raw` and `write_raw` abstract methods at the `Servo` parent. 
- [x] Define `write` and `read` functions at the parent. These functions use `read_raw` and `write_raw` defined for each protocol.
- [x] Define `replace_dictionary` at the `Servo` parent.
- [x] Define `disturbance_write_data` at the `Servo` parent.
- [x] Define `_monitoring_read_data` at the `Servo` parent.

### Tests

### DIFF
--- a/ingenialink/canopen/servo.py
+++ b/ingenialink/canopen/servo.py
@@ -81,13 +81,12 @@ class CanopenServo(Servo):
         except Exception as e:
             logger.error("Failed writing %s. Exception: %s",
                          str(reg.identifier), e)
-            error_raised = "Error writing {}".format(reg.identifier)
+            error_raised = f"Error writing {reg.identifier}"
             raise ILIOError(error_raised)
         finally:
             self._lock.release()
 
     def _read_raw(self, reg):
-        value = None
         try:
             self._lock.acquire()
             value = self.__node.sdo.upload(reg.idx, reg.subidx)
--- a/ingenialink/canopen/servo.py
+++ b/ingenialink/canopen/servo.py
@@ -27,6 +27,9 @@ class CanopenServo(Servo):
             its status, errors, faults, etc.
 
     """
+    DICTIONARY_CLASS = CanopenDictionary
+    MAX_WRITE_SIZE = CAN_MAX_WRITE_SIZE
+
     STATUS_WORD_REGISTERS = "CIA402_DRV_STATE_STATUS"
     RESTORE_COCO_ALL = "CIA301_COMMS_RESTORE_ALL"
     STORE_COCO_ALL = "CIA301_COMMS_STORE_ALL"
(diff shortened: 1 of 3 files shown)


ingenialink/canopen/servo.py
```diff
@@ -83,63 +72,29 @@ def store_parameters(self, subnode=None, sdo_timeout=3):
...
-            ILIOError: Error reading the register.
-
-        """
-        _reg = self._get_reg(reg, subnode)
-
-        access = _reg.access
-        if access == REG_ACCESS.WO:
-            raise ILAccessError('Register is Write-only')
+    def _read_raw(self, reg):
         value = None
```
MartinALL91, line 90: This line is not necessary.

ingenialink/servo.py
```diff
@@ -89,8 +92,12 @@ class Servo:
     DIST_NUMBER_SAMPLES = "DIST_CFG_SAMPLES"
     DIST_DATA = None
 
-    def __init__(self, target, servo_status_listener=False):
+    def __init__(self, target, dictionary_path=None, servo_status_listener=False):
```
MartinALL91, line 95: The docstring needs to be updated.

ingenialink/servo.py
```diff
@@ -1008,11 +1028,48 @@ def disturbance_write_data(self, channels, dtypes, data_arr):
...
+        self.disturbance_data_size = len(data)
 
-    @abstractmethod
     def _monitoring_read_data(self):
         """Read monitoring data frame."""
+        return self._read_raw(self.MONITORING_DATA)
+
+    @abstractmethod
+    def _write_raw(self, reg, data):
+        """Writes a raw bytes data to a target register.
```
MartinALL91, line 1046: Write raw bytes to a target register.

ingenialink/servo.py
```diff
@@ -1008,11 +1028,48 @@ def disturbance_write_data(self, channels, dtypes, data_arr):
...
+
+        Raises:
+            ILIOError: Error writing the register.
+
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def _read_raw(self, reg):
+        """Read raw bytes from servo.
```
MartinALL91, line 1061: Read raw bytes from a target register.

MartinALL91 (changes requested):

GullinBustin (approved):

MartinALL91 (approved):

PR #252, opened 2022-12-02
INGK-582 Avoid storing servo status information

pzinemanas:
### Description

Some information or variables of the servo, such as the status and the number of mapped registers for monitoring/disturbance, were stored at the Servo instance. Therefore, this information is simultaneously in two places (servo and lib). This could generate issues if this information is not correctly synchronized. This PR simplifies the Servo class to avoid storing this information.

Fixes # INGK-582

### Type of change

- [x] Remove `Drive.__state`
- [x] Remove `Drive.__disturbance_num_mapped_registers` and `Drive.__monitoring_num_mapped_registers`
- [x] Simplify protected attributes related to monitoring/disturbance. E.g., `__monitoring_channels_size`, `__monitoring_channels_dtype`, and `__monitoring_data` were joined into a __monitoring dictionary.
- [x] Avoid repeating the process of storing the state (reading the status_word, decoding it into a state, and saving it) several times. Instead, `self.get_state()` gets the actual state from the servo when needed.
- [x] Creates a `state_wait_change` method that is used instead of `status_word_wait_change` so we use the decoded state instead of the status_word.
- [x] The number of mapped registers f monitoring and disturbance is obtained directly from the servo instead of storing it at the Servo instance.

### Tests
- [ ] In #248 a test to check `monitoring_read_data` and `__monitoring_process_data` should be included to test the changes of this PR correctly.


ingenialink/servo.py
```diff
@@ -36,14 +36,16 @@ def __init__(self, servo):
...
         while not self.__stop:
             for subnode in range(1, self.__servo.subnodes):
                 try:
-                    status_word = self.__servo.read(
-                        self.__servo.STATUS_WORD_REGISTERS, subnode=subnode
-                    )
-                    state = self.__servo.status_word_decode(status_word)
-                    self.__servo._set_state(state, subnode=subnode)
+                    state = self.__servo.get_state(subnode)
+                    if current_state[subnode] != state:
```
MartinALL91, line 46: I think It would be clearer if the `current_state` dict is named `previous_states` and the `state`  variable is named `current_state`.

ingenialink/servo.py
```diff
@@ -36,14 +36,16 @@ def __init__(self, servo):
 
     def run(self):
         """Checks if the drive is alive by reading the status word register"""
+        current_state = {}
```
MartinALL91, line 39: You can use dictionary comprehension to simplify these lines.

ingenialink/servo.py
```diff
@@ -363,50 +350,38 @@ def enable(self, subnode=1, timeout=DEFAULT_PDS_TIMEOUT):
         """
         r = 0
```
MartinALL91, line 351: This value is not being used in this function (it is not being set elsewhere).

ingenialink/servo.py
```diff
@@ -1134,7 +1101,7 @@ def info(self):
     @property
     def monitoring_number_mapped_registers(self):
         """Get the number of mapped monitoring registers."""
-        return self.__monitoring_num_mapped_registers
+        return  self.read(self.MONITORING_NUMBER_MAPPED_REGISTERS, subnode=0)
```
MartinALL91, line 1104: Extra space

ingenialink/servo.py
```diff
@@ -1175,21 +1142,11 @@ def disturbance_data_size(self):
...
-
-        """
-        self.__disturbance_data_size = value
+        return len(self.__disturbance["data"])
 
     @property
     def disturbance_number_mapped_registers(self):
         """Get the number of mapped disturbance registers."""
-        return self.__disturbance_num_mapped_registers
+        return  self.read(self.DISTURBANCE_NUMBER_MAPPED_REGISTERS, subnode=0)
```
MartinALL91, line 1150: Extra space

ingenialink/servo.py
```diff
@@ -363,50 +350,38 @@ def enable(self, subnode=1, timeout=DEFAULT_PDS_TIMEOUT):
...
+            elif state == SERVO_STATE.NRDY:
+                cmd = constants.IL_MC_PDS_CMD_DV
+            elif state == \
+                    SERVO_STATE.DISABLED:
+                cmd = constants.IL_MC_PDS_CMD_SD
+            elif state == SERVO_STATE.RDY:
+                cmd = constants.IL_MC_PDS_CMD_SOEO
+
+            self.write(self.CONTROL_WORD_REGISTERS, cmd,
+                        subnode=subnode)
```
MartinALL91, line 377: Over-indented

ingenialink/servo.py
```diff
@@ -693,17 +672,17 @@ def disturbance_set_mapped_register(self, channel, address, subnode,
...
 
         """
-        self.__disturbance_channels_size[channel] = size
-        self.__disturbance_channels_dtype[channel] = REG_DTYPE(dtype).name
+        self.__disturbance[channel] = {
+            "size": size,
+            "dtype": REG_DTYPE(dtype).name
+        }
         data = self._monitoring_disturbance_data_to_map_register(subnode,
                                                                   address,
```
MartinALL91, line 680: Over-indented

MartinALL91 (changes requested):

ingenialink/servo.py
pzinemanas: Actually, we can use the `Servo.status` property which is exactly this dict.

WoodElv (approved):

MartinALL91 (approved):

GullinBustin (approved):

PR #278, opened 2023-02-01
INGK-602 Update docker image and stages

MartinALL91:
### Description

Now that we no longer use Ingenialink-C, the Jenkinsfile needs to be updated in order to use the new docker image (win-python-builder) to generate the python wheel and the documentation.

Fixes # INGK-602

### Type of change

- Update Jenkinsfile

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,8 +3,6 @@ def ECAT_NODE = "ecat-test"
 def ECAT_NODE_LOCK = "test_execution_lock_ecat"
 def CAN_NODE = "canopen-test"
 def CAN_NODE_LOCK = "test_execution_lock_can"
-def PYTHON_VERSIONS = ["3.6", "3.7", "3.8", "3.9"]
-def style_check = false
 
 
 pipeline {
@@ -14,7 +12,7 @@ pipeline {
             agent {
                 docker {
                     label SW_NODE
-                    image 'ingeniacontainers.azurecr.io/ingenialink-builder'
+                    image 'ingeniacontainers.azurecr.io/win-python-builder:1.0'
                 }
             }
             stages {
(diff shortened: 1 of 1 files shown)


ingenialink/utils/mcb.py
```diff
@@ -139,7 +139,7 @@ def read_mcb_data(cls, expected_address, frame):
         if expected_address != recv_add:
             raise ILWrongRegisterError(
                 f"Received address: {hex(recv_add)} does "
-                f"not match expected address: "
+                "not match expected address: "
```
WoodElv, line 142: I know there is no any variable in this line, but I think it's a little weird that this line doesn't have an f-string, when the other lines that belong to this string have it.
What do you think?

Jenkinsfile
```diff
@@ -28,36 +26,37 @@ pipeline {
...
+                             venv\\Scripts\\python.exe setup.py build sdist bdist_wheel
+                        '''
                     }
                 }
                 stage('Check formatting') {
                     steps {
                         bat """
                             cd C:\\Users\\ContainerAdministrator\\ingenialink-python
-                            py${PYTHON_VERSIONS[-1]}\\Scripts\\python.exe -m black -l 100 --check ingenialink tests
+                            venv\\Scripts\\python.exe -m black -l 100 --check ingenialink tests
```
WoodElv, line 51: I would add this flag `--experimental-string-processing` for string checking.
What do you think?

ingenialink/utils/mcb.py
MartinALL91: I think you should only use an f-string when needed, in this case, it is not.

Jenkinsfile
MartinALL91: Done

WoodElv (approved):

GullinBustin (approved):

PR #279, opened 2023-02-02
INGK-604 Add black config to toml

MartinALL91:
In this file we can store configuration settings for different tools such as black.

So we can run this command

`python -m black --check ingenialink tests`

Instead of specifying the arguments like this:

`python -m black --check --preview -l 100  ingenialink tests`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,7 +49,7 @@ pipeline {
                     steps {
                         bat """
                             cd C:\\Users\\ContainerAdministrator\\ingenialink-python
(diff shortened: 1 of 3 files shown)


WoodElv (approved):

pzinemanas (approved):

GullinBustin (changes requested):
Remove the file setup.cfg, I think it is useless and has a different line length.

MartinALL91:
> Remove the file setup.cfg, I think it is useless and has a different line length.

Done

GullinBustin (approved):

PR #282, opened 2023-02-03
INGK-605 Set correct float range

MartinALL91:
### Description

Use the correct range for floats.

Fixes # INGK-605


GullinBustin:
Jenkins Job fails

pzinemanas (approved):

GullinBustin (approved):

PR #287, opened 2023-02-24
INGK-619 Create mapped address property

MartinALL91:
### Description

In order to use monitoring/disturbance a mapping of the register address needs to be performed. Create a property in the Register class to get this mapped address.

Fixes # INGK-619

### Type of change

- Added a mapped_address property to EthernetRegister and CanopenRegister

### Tests
- Added test_register_mapped_address

### DIFF
--- a/ingenialink/canopen/register.py
+++ b/ingenialink/canopen/register.py
@@ -82,3 +82,8 @@ class CanopenRegister(Register):
     def subidx(self):
         """int: Register subindex."""
         return self.__subidx
+
+    @property
+    def mapped_address(self):
+        """int: Register mapped address used for monitoring/disturbance."""
+        return self.idx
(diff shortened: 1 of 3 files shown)


ingenialink/canopen/register.py
```diff
@@ -82,3 +82,8 @@ def idx(self):
     def subidx(self):
         """int: Register subindex."""
         return self.__subidx
+
+    @property
+    def mapped_address(self):
+        """int: Register mapped address used for monitoring/disturbance."""
+        return self.idx
```
GullinBustin, line 89: the mapped address does not change with the subnode?

MartinALL91: No, it's the same as the idx

GullinBustin (approved):

pzinemanas (approved):

PR #309, opened 2023-08-11
Ingk 666 mutex should not be created inside in a thread

WoodElv:
### Description

The issue  was there were no mutual exclusion for some shared resources. The reason is mutexes were created inside per each thread, and it seems mutexes in python need to be created outside them, because before locks were used just by their own thread.

Fixes # (issue)

### Type of change

- [x ] Locks are created outside threads

### Code formatting

- [ ] Use black package to format the code: `black -l 100 ingenialink tests`. It is recommended to configure the code editor to automatically format the code using black with a max length line of 100.

### DIFF
--- a/ingenialink/poller.py
+++ b/ingenialink/poller.py
@@ -38,6 +38,7 @@ class PollerTimer:
         if self.thread.is_alive():
             self.thread.join()
 
+lock_poller = Lock()
 
 class Poller:
     """Register poller for CANOpen/Ethernet communications.
@@ -60,7 +61,6 @@ class Poller:
         self.__running = False
         self.__mappings = []
         self.__mappings_enabled = []
-        self.__lock = Lock()
         self._reset_acq()
 
     def start(self):
@@ -198,7 +198,7 @@ class Poller:
         # Obtain current time
         t = delta.total_seconds()
 
-        self.__lock.acquire()
+        lock_poller.acquire()
         # Acquire all configured channels
         if self.__samples_count >= self.__sz:
             self.__samples_lost = True
@@ -221,7 +221,7 @@ class Poller:
             # Increment samples count
             self.__samples_count += 1
 
-        self.__lock.release()
+        lock_poller.release()
 
     @property
     def data(self):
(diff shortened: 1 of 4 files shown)


ingenialink/poller.py
```diff
@@ -38,6 +38,7 @@ def cancel(self):
         if self.thread.is_alive():
             self.thread.join()
 
+lock_poller = Lock()
```
jfernandez-ingenia, line 41: The poller should use the same lock as the active communication. If you create two different locks, the exclusion mechanism is not going to work.

ingenialink/poller.py
```diff
@@ -198,7 +198,7 @@ def _acquire_callback_poller_data(self):
         # Obtain current time
         t = delta.total_seconds()
 
-        self.__lock.acquire()
+        lock_poller.acquire()
```
MartinALL91, line 201: I'm not sure if we need to acquire the lock here. The lock will be acquired [here ](https://github.com/ingeniamc/ingenialink-python/pull/309/files#diff-828b231da05ac56f01ef705bb7514038b48b067402fba9c7d4fd4e409ac4577aL217) when reading a register.

ingenialink/poller.py
WoodElv: As @MartinALL91 comments [here](https://github.com/ingeniamc/ingenialink-python/pull/309/files#r1291041168), the lock is not needed for the reading. It should only affects that the sample that a channel is reading will late milliseconds more. In fact, threads are using the same lock finally, which are the locks of each servo.

However, I would keep this lock because of [this](https://github.com/ingeniamc/ingenialink-python/pull/309/files#diff-828b231da05ac56f01ef705bb7514038b48b067402fba9c7d4fd4e409ac4577aR246). In this case, as it's another shared resource, I think it makes sense we use another lock.

What do you think?

MartinALL91 (approved):

ingenialink/poller.py
WoodElv: I think we can delete the lock here.

ingenialink/poller.py
```diff
@@ -221,7 +221,7 @@ def _acquire_callback_poller_data(self):
             # Increment samples count
             self.__samples_count += 1
 
-        self.__lock.release()
+        lock_poller.release()
```
WoodElv, line 224: And delete it here as well

GullinBustin:
I think this PR has no sense anymore.

PR #321, opened 2023-10-19
INGK-315 Modify status network property in case of multislave

MartinALL91:
### Description

Adapt CANopen's Network status listener to support multi-drives.

Fixes #INGK-315

### Type of change

-  The subscribe_to_status function takes the node ID as argument.
- The start and stop status listener functions no longer require the servo as argument.

### Tests
- Connect to two CANopen drives, subscribe to their status, and check that the callback functions are notified of the network status changes.

### DIFF
--- a/ingenialink/canopen/network.py
+++ b/ingenialink/canopen/network.py
@@ -3,6 +3,7 @@ from enum import Enum
 from time import sleep
 from threading import Thread
 from typing import Optional, Callable, Union
+from collections import defaultdict
 
 from .register import CanopenRegister
 from ingenialink.utils.mcb import MCB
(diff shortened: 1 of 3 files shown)


pzinemanas (approved):

ingenialink/canopen/network.py
```diff
@@ -119,34 +120,35 @@ class NetStatusListener(Thread):
...
     def run(self):
+        timestamps = {}
         while not self.__stop:
-            if self.__timestamp == self.node.nmt.timestamp:
-                if self.__state != NET_STATE.DISCONNECTED:
-                    self.__network.status = NET_STATE.DISCONNECTED
-                    self.__state = NET_STATE.DISCONNECTED
-                    self.__network._notify_status(NET_DEV_EVT.REMOVED)
-                else:
+            for node_id, node in list(self.__network._connection.nodes.items()):
```
GullinBustin, line 134: Why you use list instead of use items() output directly?

MartinALL91: This is what I mentioned in the daily, this way the node list is forced to reload.

GullinBustin (approved):

PR #322, opened 2023-10-24
Bugfix/ingk 680 disturbance create data chunks raise overflow error if values do not fit in registers

MartinALL91:
### Description
Raise ILValueError when the disturbance data does not fit in a register.

Fixes #INGK-680

### Type of change

- Raise ILValueError when the disturbance data does not fit in a register.

### Tests
- Create a disturbance signal that cannot fit in the register dtype (for example the signal contains negative values and the register dtype is unsigned).
- Try to write the disturbance data.
- Check that the ILValueError exception is raised.

### DIFF
--- a/ingenialink/servo.py
+++ b/ingenialink/servo.py
@@ -12,6 +12,7 @@ from ingenialink.exceptions import (
     ILStateError,
     ILAccessError,
     ILTimeoutError,
+    ILValueError,
 )
 from ingenialink.register import Register
 from ingenialink.utils._utils import (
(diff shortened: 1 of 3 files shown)


pzinemanas (approved):

ingenialink/servo.py
```diff
@@ -997,9 +998,12 @@ def disturbance_write_data(self, channels, dtypes, data_arr):
...
         """
-        data, chunks = self._disturbance_create_data_chunks(
-            channels, dtypes, data_arr, self.MAX_WRITE_SIZE
-        )
+        try:
+            data, chunks = self._disturbance_create_data_chunks(
+                channels, dtypes, data_arr, self.MAX_WRITE_SIZE
+            )
+        except OverflowError as e:
+            raise ILValueError(f"Disturbance data cannot be written. Reason: {e}")
```
GullinBustin, line 1006: You can use `raise ILValueError(f"Disturbance data cannot be written.") from e`

GullinBustin (changes requested):

ingenialink/servo.py
MartinALL91: Done

GullinBustin (approved):

PR #339, opened 2023-11-28
INGK-729 Handle EMCY messages

MartinALL91:
### Description

Raise an exception if an emergency message is received.

Fixes #INGK-729

### Type of change

- Add the _get_emergency_description function

### DIFF
--- a/ingenialink/ethercat/servo.py
+++ b/ingenialink/ethercat/servo.py
@@ -84,7 +84,7 @@ class EthercatServo(Servo):
         try:
             self.__slave.sdo_write(reg.idx, reg.subidx, data, complete_access)
             self._check_working_counter()
-        except (SdoError, MailboxError, PacketError, ILIOError, Emergency) as e:
+        except (SdoError, MailboxError, PacketError, ILIOError) as e:
             raise ILIOError(f"Error writing {reg.identifier}. Reason: {e}") from e
         except Emergency as e:
             error_description = self._get_emergency_description(e.error_code)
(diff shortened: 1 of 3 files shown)


ingenialink/ethercat/servo.py
```diff
@@ -63,24 +63,66 @@ def _read_raw(  # type: ignore [override]
         self, reg: EthercatRegister, buffer_size: int = 0, complete_access: bool = False
     ) -> bytes:
         self._lock.acquire()
+        try:
+            value = self._read_raw_attempt(reg, buffer_size, complete_access)
+        except Emergency:
+            logger.info("An emergency message was received. Retrying read operation.")
```
GullinBustin, line 69: Why do you retry? If an error appears, it should be caught and raised as an ILError. What kind of errors are Emergency errors?

WoodElv (approved):

ingenialink/ethercat/servo.py
MartinALL91: Over voltages and so on. We handle these kinds of messages [differently](https://github.com/ingeniamc/ingeniamotion/blob/master/ingeniamotion/errors.py#L136). In any case, maybe what we should do is to handle them as is its done with [CANopen](https://github.com/ingeniamc/ingenialink-python/blob/master/ingenialink/canopen/servo.py#L114). For now, I think we should just ignore them as they interefe with the correct functioning of our code.

ingenialink/ethercat/servo.py
```diff
@@ -1,4 +1,4 @@
-from pysoem import CdefSlave, SdoError, MailboxError, PacketError  # type: ignore
+from pysoem import CdefSlave, SdoError, MailboxError, PacketError, Emergency, pysoem  # type: ignore
```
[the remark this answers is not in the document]

pzinemanas (approved):

GullinBustin (approved):

PR #343, opened 2023-12-21
INGK-766 Make pysoem import optional

GullinBustin:
### Description

If WinPcap is not installed, do not raise an exception on import, but on ethercat network init

Fixes # INGK-766

### Type of change

- [x] Add try-except on pysoem import in ethercat network and servo
- [x] Raise an exception on ethercat network and servo if pysoem couldn't be imported

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### DIFF
--- a/ingenialink/ethercat/network.py
+++ b/ingenialink/ethercat/network.py
@@ -12,7 +12,9 @@ import ingenialogger
 
 try:
     import pysoem  # type: ignore
-except ImportError:
+except ImportError as e:
+    if isinstance(e, ModuleNotFoundError):
+        raise e
     pysoem = None
 
 from ingenialink.network import Network, NET_PROT, NET_STATE, NET_DEV_EVT
@@ -29,6 +31,9 @@ class NetStatusListener(Thread):
 
     Args:
         network: Network instance of the EtherCAT communication.
+
+    Raises:
+        ImportError: WinPcap is not installed
 
     """
 
--- a/ingenialink/ethercat/network.py
+++ b/ingenialink/ethercat/network.py
@@ -12,10 +12,9 @@ import ingenialogger
 
 try:
     import pysoem  # type: ignore
-except ImportError as e:
-    if isinstance(e, ModuleNotFoundError):
-        raise e
+except ImportError as ex:
     pysoem = None
+    pysoem_import_error = ex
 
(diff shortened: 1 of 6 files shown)


WoodElv (approved):

MartinALL91 (approved):

ingenialink/ethercat/network.py
```diff
@@ -76,6 +80,8 @@ class EthercatNetwork(Network):
     def __init__(
         self, interface_name: str, connection_timeout: float = DEFAULT_ECAT_CONNECTION_TIMEOUT
     ):
+        if not pysoem:
+            raise ImportError("DLLs required not found: Please install WinPcap and restart")
```
polfeliu, line 84: This could be masking other import errors.

Something like this could be done:

```
try:
    import pysoem  # type: ignore
except ImportError as ex:
    pysoem = None
    pysoem_import_error = ex
```

```
if not pysoem:
    raise pysoem_import_error ?
```

Or saving the message instead of the exception object

ingenialink/ethercat/network.py
```diff
@@ -9,7 +9,13 @@
 from threading import Thread
 
 import ingenialogger
-import pysoem  # type: ignore
+
+try:
+    import pysoem  # type: ignore
+except ImportError as e:
+    if isinstance(e, ModuleNotFoundError):
```
polfeliu, line 16: Is this the only thing that should not be masked? I have no idea what errors can happen during an import

ingenialink/ethercat/servo.py
```diff
@@ -50,11 +63,13 @@ class EthercatServo(Servo):
 
     def __init__(
         self,
-        slave: CdefSlave,
+        slave: "CdefSlave",
```
polfeliu, line 66: I've never seen double quotes for forward reference of type:
https://peps.python.org/pep-0484/#runtime-or-type-checking

GullinBustin: black changes it automatically to double quotes

ingenialink/ethercat/network.py
GullinBustin: Mmm, I like it. With this change, we can not say explicitly that WinPcap is needed, but we will not mask any error. I prefer it.

PR #362, opened 2024-01-26
Ingk 799 implement processdata function

GullinBustin:
### Description

Implement porcessdata function and update some PDO functions

Fixes # (INGK-799 )

### Type of change

- [x] Update PDO example
- [x] Create send_receive_processdata function
- [x] Check RPDOs initial data in start_pdos
- [x] Update set_pdo_map_to_slave function
- [x] Update map_tpdos and map_rpdos functions

### Tests
- [x] Run tests.

### Documentation

### Code formatting


ingenialink/ethercat/network.py
```diff
@@ -200,19 +200,21 @@ def start_pdos(self) -> None:
             ILStateError: If slaves can not reach SafeOp state
 
         """
+        op_servo_list = [servo for servo in self.servos if servo._rpdo_maps or servo._tpdo_maps]
+        try:
+            for servo in op_servo_list:
+                for rpdo_map in servo._rpdo_maps:
+                    rpdo_map.get_item_bytes()
+        except ILError as e:
+            raise ILError("RPDOs initial value should be set before start PDOs") from e
```
MartinALL91, line 209: The RPDO values should be set before starting the PDO exchange process.

ingenialink/ethercat/network.py
```diff
@@ -200,19 +200,21 @@ def start_pdos(self) -> None:
...
         self._ecat_master.config_map()
         self._ecat_master.state = pysoem.SAFEOP_STATE
-        op_servo_list = [servo for servo in self.servos if servo._rpdo_maps or servo._tpdo_maps]
         if not op_servo_list:
```
MartinALL91, line 212: This should be checked immediately after creating the `op_servo_list`

ingenialink/ethercat/network.py
```diff
@@ -200,19 +200,21 @@ def start_pdos(self) -> None:
...
             logger.warning("No drives has PDO mapping")
```
MartinALL91, line 213: There are no PDOs assigned to any connected slave.

ingenialink/ethercat/network.py
```diff
@@ -226,6 +228,16 @@ def stop_pdos(self) -> None:
         if not self._change_nodes_state(op_servo_list, pysoem.PREOP_STATE):
             logger.warning("Drive can not reach PreOp state")
 
+    def send_receive_processdata(self) -> None:
```
MartinALL91, line 231: Missing docstring

ingenialink/ethercat/servo.py
```diff
@@ -258,7 +254,11 @@ def _get_emergency_description(self, error_code: int) -> Optional[str]:
         return error_description
 
     def set_pdo_map_to_slave(self, rpdo_maps: List[RPDOMap], tpdo_maps: List[TPDOMap]) -> None:
```
MartinALL91, line 256: Missing docstring

ingenialink/pdo.py
```diff
@@ -306,107 +321,80 @@ def reset_tpdo_mapping(self) -> None:
...
         """
-        if len(self._rpdo_maps) + 1 > self.AVAILABLE_PDOS:
-            raise ILError("Could not add a new RPDO map, there are no available PDOs.")
-        self._rpdo_maps.append(rpdo_map)
-        map_index = len(self._rpdo_maps) - 1
-
+        if len(self._rpdo_maps) > self.AVAILABLE_PDOS:
+            raise ILError(
+                f"Could not map the RPDO maps, received {len(self._rpdo_maps)} PDOs and only"
+                " {self.AVAILABLE_PDOS} are available"
```
MartinALL91, line 334: This is a regular string, should be an f-string.

ingenialink/pdo.py
```diff
@@ -306,107 +321,80 @@ def reset_tpdo_mapping(self) -> None:
...
 
         Raises:
             ILError: If there are no available PDOs.
         """
-        if len(self._rpdo_maps) + 1 > self.AVAILABLE_PDOS:
-            raise ILError("Could not add a new RPDO map, there are no available PDOs.")
-        self._rpdo_maps.append(rpdo_map)
-        map_index = len(self._rpdo_maps) - 1
-
+        if len(self._rpdo_maps) > self.AVAILABLE_PDOS:
```
MartinALL91, line 331: I think you should create a variable to store the number of RPDO maps.

ingenialink/pdo.py
```diff
@@ -306,107 +321,80 @@ def reset_tpdo_mapping(self) -> None:
...
         self.write(self.RPDO_ASSIGN_REGISTER_SUB_IDX_0, len(self._rpdo_maps))
-        self.write(self.RPDO_MAP_REGISTER_SUB_IDX_0[map_index], len(rpdo_map.items))
-        self.write(
-            self.RPDO_MAP_REGISTER_SUB_IDX_1[map_index],
-            rpdo_map.items_mapping.decode("utf-8"),
-            complete_access=True,
-        )
+        custom_map_index = 0
+        rpdo_assigns = b""
+        for rpdo_map in self._rpdo_maps:
```
MartinALL91, line 339: You could do:
`for custom_map_index, rpdo_map in enumerate(self._rpdo_maps):`

ingenialink/pdo.py
```diff
@@ -306,107 +321,80 @@ def reset_tpdo_mapping(self) -> None:
...
 
         Raises:
             ILError: If there are no available PDOs.
         """
-        if len(self._tpdo_maps) + 1 > self.AVAILABLE_PDOS:
-            raise ILError("Could not add a new TPDO map, there are no available PDOs.")
-        self._tpdo_maps.append(tpdo_map)
-        map_index = len(self._tpdo_maps) - 1
-
+        if len(self._tpdo_maps) > self.AVAILABLE_PDOS:
```
MartinALL91, line 364: Why are the `_tpdo_maps` and `_rpdo_maps` lists, when we can only have a single PDO map?

MartinALL91 (changes requested):

ingenialink/ethercat/network.py
```diff
@@ -224,6 +228,16 @@ def stop_pdos(self) -> None:
...
+        for servo in self.servos:
+            servo.process_pdo_inputs()
+        self._ecat_master.send_processdata()
+        processdata_wkc = self._ecat_master.receive_processdata(
+            timeout=self.ECAT_PROCESSDATA_TIMEOUT_NS
```
polfeliu, line 236: I think this timeout should be exposed to the user with a default

ingenialink/ethercat/servo.py
```diff
@@ -71,17 +71,15 @@ class EthercatServo(PDOServo):
...
-            identifier="RPDO_ASSIGN_REGISTER",
-            units="",
-            subnode=0,
-            idx=0x1C12,
-            subidx=0x01,
-            dtype=REG_DTYPE.S32,
-            access=REG_ACCESS.RW,
-        )
-    ]
+    RPDO_ASSIGN_REGISTER_SUB_IDX_1 = EthercatRegister(
```
polfeliu, line 74: Why is this not a list anymore?
If by using complete access we are using more than the declared subidx here this is really confusing.
Maybe it's time we create a "SubindexableEthercatRegister" class that represents registers that have multiple subindexes

ingenialink/pdo.py
polfeliu: Why have more complexity and maintenance?
If it's a worry about performance read this: https://stackoverflow.com/questions/699177/python-do-python-lists-keep-a-count-for-len-or-does-it-count-for-each-call

ingenialink/pdo.py
polfeliu: in the future we will have support for multiple pdo maps like many ethercat devices

ingenialink/ethercat/servo.py
GullinBustin: I think it makes no sense that it was a list, we only have one RPDO_ASSIGN_REGISTER_SUB_IDX_1 and it can't be more. If you want a list element for each subindex then the name should be RPDO_ASSIGN_REGISTER.
About "SubindexableEthercatRegister", yes, probably we will need something like that we can use the new dictionary definition to think about it.

ingenialink/pdo.py
GullinBustin: It is not the same, if a rpdo_mao is read-only, then custom_map_index is not increased

ingenialink/pdo.py
GullinBustin: It will change for now FW, that should be defined in dicctionaty

ingenialink/ethercat/servo.py
GullinBustin: It is defined in parent class

ingenialink/pdo.py
GullinBustin: This function has 3 steps:

1. Set RPDO assign register size.
2. Fill custom RPDOs to RPOD map registers. That only must be done if it is not read-only PDO.
3. Set all the RPDO maps to RPDO assign register

GullinBustin: Finally I divide it in two functions

ingenialink/ethercat/servo.py
polfeliu: Well it did make sense if each subidx had a definition. 
Since we only contemplated one pdo map assignment a list with one item was the correct choice.
Now this change is confusing, I would add a comment explaining that this register is representing all the subindexes by using complete access later, and that will be replaced with the "SubindexableEthercatRegister" when it's implemented.
And probably remove SUB_IDX_1 from the name

MartinALL91 (approved):

ingenialink/ethercat/servo.py
GullinBustin: Technically it is just the sub-index 1, it is indeed used for complete access, but it does not affect this variable.

pzinemanas (approved):

PR #376, opened 2024-02-16
INGK-835 Raise exception if unexpected working counter is received

MartinALL91:
### Description

If the working counter value is not expected (different than [NO_FRAME](https://github.com/ingeniamc/ingenialink-python/blob/develop/ingenialink/ethercat/servo.py#L54) and [TIMEOUT](https://github.com/ingeniamc/ingenialink-python/blob/develop/ingenialink/ethercat/servo.py#L53)), no exception is raised.

Fixes # INGK-835

### Type of change

- Raise exception if an unexpected working counter value is received

### Tests
- [x] Run tests.

### Documentation

### Code formatting

### DIFF
--- a/ingenialink/ethercat/servo.py
+++ b/ingenialink/ethercat/servo.py
@@ -253,8 +253,12 @@ class EthercatServo(PDOServo):
         elif isinstance(exception, pysoem.WkcError):
             if exception.wkc == self.NOFRAME_WORKING_COUNTER:
                 raise ILIOError(f"Error {operation_msg.value} data: No frame.") from exception
-            if exception.wkc == self.TIMEOUT_WORKING_COUNTER:
+            elif exception.wkc == self.TIMEOUT_WORKING_COUNTER:
                 raise ILTimeoutError(f"Timeout {operation_msg.value} data.") from exception
+            else:
+                raise ILIOError(
+                    f"Error {operation_msg.value} data. Wrong working counter: {exception.wkc}"
+                ) from exception
 
     def _monitoring_read_data(self) -> bytes:  # type: ignore [override]
         """Read monitoring data frame."""


pzinemanas (approved):

ingenialink/ethercat/servo.py
```diff
@@ -251,10 +253,21 @@ def _handle_sdo_exception(
                 f"Error {operation_msg.value} {reg.identifier}. Reason: {exception}"
             ) from exception
         elif isinstance(exception, pysoem.WkcError):
-            if exception.wkc == self.NOFRAME_WORKING_COUNTER:
-                raise ILIOError(f"Error {operation_msg.value} data: No frame.") from exception
-            if exception.wkc == self.TIMEOUT_WORKING_COUNTER:
-                raise ILTimeoutError(f"Timeout {operation_msg.value} data.") from exception
+            default_error_msg = f"Error {operation_msg.value} data"
+            wkc_exceptions: Dict[int, Exception] = {
```
WoodElv, line 257: Must you use the generic Exception type? I think you could use:
`Union[ILIOError, ILTimeoutError]`

WoodElv (changes requested):

ingenialink/ethercat/servo.py

WoodElv (approved):

ingenialink/ethercat/servo.py
```diff
@@ -50,8 +50,9 @@ class EthercatServo(PDOServo):
     DICTIONARY_CLASS = EthercatDictionary
     MAX_WRITE_SIZE = CAN_MAX_WRITE_SIZE
 
+    DEFAULT_WORKING_COUNTER = 0
```
GullinBustin, line 53: I think default is not a good name if it means that something fail

ingenialink/ethercat/servo.py
```diff
@@ -251,10 +253,21 @@ def _handle_sdo_exception(
                 f"Error {operation_msg.value} {reg.identifier}. Reason: {exception}"
             ) from exception
         elif isinstance(exception, pysoem.WkcError):
-            if exception.wkc == self.NOFRAME_WORKING_COUNTER:
-                raise ILIOError(f"Error {operation_msg.value} data: No frame.") from exception
-            if exception.wkc == self.TIMEOUT_WORKING_COUNTER:
```
GullinBustin, line 256: I think write different if is more clear than the new way

GullinBustin (changes requested):

ingenialink/ethercat/servo.py
MartinALL91: When the master sends an EtherCAT frame it sets the working counter to zero, If this value is unchanged in the reply then it means that something went wrong. I think that the name is correct. The error message is also clear.

GullinBustin: If someone doesn't know what working counter 0 means and reads this name will think that 0 is good. We must write explicitly that 0 means error or fail in this case.

GullinBustin (approved):

PR #384, opened 2024-03-18
INGK-849 SafeOp cannot be reached after calling config map

MartinALL91:
### Description

Follow the right configuration sequence to change the PDO mapping.

Fixes # INGK-849

### Type of change

- After the PDO exchange is stopped, transition back to Init and reconfigure all the slaves.
- Transition back to Pre-Op state.

### Tests
- [x] Test the re-configuration of the PDOs.

### Code formatting

### DIFF
--- a/tests/ethercat/test_pdo.py
+++ b/tests/ethercat/test_pdo.py
@@ -273,6 +273,35 @@ def connect_to_all_slave(pytestconfig):
         net.disconnect_from_slave(servo)
 
 
+def create_pdo_maps(servo, rpdo_registers, tpdo_registers):
+    rpdo_map = RPDOMap()
+    tpdo_map = TPDOMap()
+    for tpdo_register in tpdo_registers:
+        register = servo.dictionary.registers(SUBNODE)[tpdo_register]
+        tpdo_map.add_registers(register)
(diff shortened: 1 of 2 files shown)


pzinemanas (approved):

tests/ethercat/test_pdo.py
```diff
@@ -321,12 +330,15 @@ def test_start_stop_pdo(connect_to_all_slave):
...
-    # Check that PDOs can be re-started with the same configuration
-    net.start_pdos()
-    start_time = time.time()
-    while time.time() < start_time + timeout:
-        net.send_receive_processdata()
-    net.stop_pdos()
+    # Re-configure the PDOs and re-start the PDO exchange
+    for servo in servos:
+        servo.remove_rpdo_map(rpdo_map_index=0)
+        servo.remove_tpdo_map(tpdo_map_index=0)
```
GullinBustin, line 336: Why do you reconfigure the PDOs? Do you test that you can simply stop and start PDO with the same configuration?

MartinALL91: Because there is no difference if we use the previous or a new configuration.

GullinBustin (approved):

PR #390, opened 2024-03-27
INGK-467 Add support to socketcan (Linux)

pzinemanas:
### Description

In this PR the support to socketcan device is added. So the canopen communication can be used in Linux platforms.

Fixes # INGK-467

### Type of change

- [x] Add support to socketcan.
- [x] Fix import errors in linux.

### Tests
- [x] Test connection and SDOs.
- [x] Test load FW.
- [x] Run unit tests in a Linux machine.

### Documentation

### Code formatting


ingenialink/canopen/network.py
```diff
@@ -354,6 +362,8 @@ def _setup_connection(self) -> None:
         """Creates a network interface object establishing an empty connection
         with all the network attributes already specified."""
         if self._connection is None:
+            if not RUNNING_ON_WINDOWS and self.__device != CAN_DEVICE.SOCKETCAN.value:
```
MartinALL91, line 365: I would move this check to the constructor.

ingenialink/canopen/network.py
```diff
@@ -8,10 +8,16 @@
 from time import sleep
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
+RUNNING_ON_WINDOWS = os.name == "nt"
```
MartinALL91, line 11: We could use the [system](https://docs.python.org/3/library/platform.html#platform.system) method to check the OS.

ingenialink/canopen/network.py
```diff
@@ -631,11 +641,16 @@ def __program_control_to_stop(
         logger.info("Flashing firmware")
         with contextlib.suppress(ILError):
             servo.write(PROG_STAT_1, PROG_CTRL_STATE_STOP, subnode=0)
-        try:
+        if RUNNING_ON_WINDOWS:
```
MartinALL91, line 644: I don't think this is necessary.

MartinALL91 (changes requested):

ingenialink/canopen/network.py
MartinALL91: We should check if the OS is linux.

ingenialink/canopen/network.py
pzinemanas: The thing is that the VCIError would not be defined when running on windows.

ingenialink/canopen/network.py
pzinemanas: Done [9d6f708](https://github.com/ingeniamc/ingenialink-python/pull/390/commits/9d6f708707fbcf386670c5534ef19ed948f7e0bb)

ingenialink/canopen/network.py
pzinemanas: Done [9d6f708](https://github.com/ingeniamc/ingenialink-python/pull/390/commits/9d6f708707fbcf386670c5534ef19ed948f7e0bb)

ingenialink/canopen/network.py
```diff
@@ -185,6 +194,8 @@ def __init__(
         baudrate: CAN_BAUDRATE = CAN_BAUDRATE.Baudrate_1M,
     ):
         super(CanopenNetwork, self).__init__()
+        if not platform.system() == "Linux" and device != CAN_DEVICE.SOCKETCAN.value:
+            raise ILError("In Linux machines, only the SOCKETCAN device can be used.")
```
GullinBustin, line 198: Why? Technically the other drivers can be used too.

GullinBustin (changes requested):

ingenialink/canopen/network.py
pzinemanas: That's right. I'll remove this error.

MartinALL91 (approved):

PR #404, opened 2024-04-25
INGK-874 Fix FoE when there are blank spaces in the bin path

pzinemanas:
### Description

This PR fixes the bug when there are spaces in the path to FoE binary in Windows.

Fixes # INGK-874 

### Type of change

- [x] Back to the previous way to pass the attributes to the FoE binary in Windows.

### Tests
- [x] Test it in ML3.

### Documentation

### Code formatting

### Others

### DIFF
--- a/ingenialink/ethercat/network.py
+++ b/ingenialink/ethercat/network.py
@@ -450,12 +450,20 @@ class EthercatNetwork(Network):
             except subprocess.CalledProcessError as e:
                 raise ILFirmwareLoadError("Could not change the FoE binary permissions.") from e
         try:
-            subprocess.run(
-                f"{exec_path} {self.interface_name} {slave_id} {fw_file}",
-                check=True,
-                shell=True,
-                encoding="utf-8",
-            )
+            if sys_name == "linux":
+                subprocess.run(
+                    f"{exec_path} {self.interface_name} {slave_id} {fw_file}",
+                    check=True,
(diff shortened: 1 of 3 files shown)


ingenialink/ethercat/network.py
```diff
@@ -450,12 +450,20 @@ def load_firmware(self, fw_file: str, slave_id: int = 1) -> None:  # type: ignor
             except subprocess.CalledProcessError as e:
                 raise ILFirmwareLoadError("Could not change the FoE binary permissions.") from e
         try:
-            subprocess.run(
```
MartinALL91, line 453: This block of code can be simplified, like so:
```
if sys_name == "linux":
    try:
        subprocess.run(
            f"chmod 777 {exec_path}",
            check=True,
            shell=True,
            encoding="utf-8",
        )
    except subprocess.CalledProcessError as e:
        raise ILFirmwareLoadError("Could not change the FoE binary permissions.") from e
    args = f"{exec_path} {self.interface_name} {slave_id} {fw_file}"
else:
    args = [exec_path, self.interface_name, f"{slave_id}", fw_file]
try:
    subprocess.run(
        args,
        check=True,
        shell=True,
        encoding="utf-8",
    )
```

MartinALL91 (approved):

WoodElv (approved):

PR #428, opened 2024-06-03
INGK-804 add typing to examples

mjedieckmann:
### Description

Add type checking to all the files in the example-folder.
The tox type check (Jenkins) will also check these files now.

[Issue](https://novantamotion.atlassian.net/browse/INGK-804)

### Tests
- [ ] Run Jenkins type checking.
- [x] Check manually using tox -e type

### Others

- [x] Update changelog

### DIFF
--- a/examples/canopen/can_disturbance.py
+++ b/examples/canopen/can_disturbance.py
@@ -1,11 +1,12 @@
-import sys
 import math
+import sys
+from typing import List, Union, cast
 
-from ingenialink.canopen.network import CanopenNetwork, CAN_DEVICE, CAN_BAUDRATE
+from ingenialink.canopen.network import CAN_BAUDRATE, CAN_DEVICE, CanopenNetwork
 from ingenialink.register import REG_DTYPE
 
 
-def disturbance_example():
+def disturbance_example() -> None:
     # Frequency divider to set disturbance frequency
     divider = 100
     # Calculate time between disturbance samples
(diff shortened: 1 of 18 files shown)


examples/ethercat/ecat_load_firmware.py
```diff
@@ -3,10 +3,10 @@
 from ingenialink.ethercat.network import EthercatNetwork
 
 
-def ecat_load_fw():
+def ecat_load_fw() -> None:
     net = EthercatNetwork("\\Device\\NPF_{192D1D2F-C684-467D-A637-EC07BD434A63}")
 
-    net.load_firmware(fw_file='../../resources/firmware/cap-net-e_0.7.1.lfu')
+    net.load_firmware(fw_file="../../resources/firmware/cap-net-e_0.7.1.lfu", boot_in_app=False)
```
MartinALL91, line 9: This could be dangerous. I think we should add a comment explaining how to determine the correct value for this parameter.

examples/canopen/can_disturbance.py
```diff
@@ -19,11 +20,44 @@ def disturbance_example():
     # A = signal_amplitude (Amplitude)
     # t = sample_period*i (time)
     # w = signal_frequency*2*math.pi (angular frequency)
-    data_pos = [int(1000*signal_amplitude * math.sin(sample_period*i * signal_frequency * 2*math.pi)) for i in range(n_samples)]
-    data_vel = [signal_amplitude * math.sin(sample_period*i * signal_frequency * 2*math.pi) for i in range(n_samples)]
-    data_curr_q = [signal_amplitude * math.sin(sample_period*i * signal_frequency * 2*math.pi - math.pi/2) for i in range(n_samples)]
-    data_curr_d = [signal_amplitude * math.sin(sample_period*i * signal_frequency * 2*math.pi + math.pi/2) for i in range(n_samples)]
-    data_positioning_opt = [int(abs(500*signal_amplitude * math.sin(sample_period*i * signal_frequency * 2*math.pi + math.pi))) for i in range(n_samples)]
+    data_pos = cast(
+        List[Union[int, float]],
```
polfeliu, line 24: I don't understand why a explicit cast is necessary here

mjedieckmann: Mypy will complain if we don't since having a list of numbers that can be int or float is essentially too uncertain for the type check to work.
I'm not sure I entirely understand how this works or why it's a problem, but the explicit cast makes mypy happy enough.
More info:
https://stackoverflow.com/questions/43957034/specifying-a-type-to-be-a-list-of-numbers-ints-and-or-floats

examples/ethercat/ecat_load_firmware.py
mjedieckmann: I've instead changed the actual function to have "False" as a default if the parameter is not provided since the use of the parameter is described in the docstring of the function.

ingenialink/ethercat/network.py
```diff
@@ -409,7 +409,7 @@ def stop_status_listener(self) -> None:  # type: ignore [override]
             self.__listener_net_status.join()
         self.__listener_net_status = None
 
-    def load_firmware(self, fw_file: str, boot_in_app: bool, slave_id: int = 1) -> None:  # type: ignore [override]
+    def load_firmware(self, fw_file: str, boot_in_app: bool = False, slave_id: int = 1) -> None:  # type: ignore [override]
```
mjedieckmann, line 412: added default

MartinALL91 (approved):

examples/canopen/can_disturbance.py
polfeliu: Oh, I see the problem.
The method argument disturbance_write_data is not correctly typed.
It accepts a int/float covariant list, and since the list is not immutable the variance could be change by the method and force a new internal type.
I think we should review this but maybe not in this issue. Forcing the user to use so many verbose casts seems ridiculous to me

GullinBustin (approved):

PR #447, opened 2024-06-19
INGK-893 Recover EEPROM write functionality for manufacturing

MartinALL91:
### Description

Add methods to read and write to the ESC EEPROM.

Fixes # INGK-893

### Type of change

- Add the read_eeprom method.
- Add the write_eeprom method.

### Tests
- [x] Run tests.

### Documentation

### Others

### DIFF
--- a/ingenialink/ethercat/servo.py
+++ b/ingenialink/ethercat/servo.py
@@ -59,6 +59,8 @@ class EthercatServo(PDOServo):
 
     DEFAULT_STORE_RECOVERY_TIMEOUT = 1
 
+    DEFAULT_EEPROM_OPERATION_TIMEOUT_uS = 200_000
+
     def __init__(
         self,
         slave: "CdefSlave",
@@ -319,6 +321,39 @@ class EthercatServo(PDOServo):
         self.slave.set_watchdog(
             self.ETHERCAT_PDO_WATCHDOG, self.SECONDS_TO_MS_CONVERSION_FACTOR * timeout
         )
+
+    def read_eeprom(
+        self, address: int, timeout: int = DEFAULT_EEPROM_OPERATION_TIMEOUT_uS
+    ) -> bytes:
+        """Read from the ESC EEPROM.
+
+        Args:
+            address: EEPROM address to be read.
+            timeout: Operation timeout (microseconds). By default, 200.000 us.
+
+        Returns:
+            EEPROM data. The read data is 2 words (4 bytes).
+
+        """
+        data = self.slave.eeprom_read(address, timeout)
+        if not isinstance(data, bytes):
+            raise ValueError(
+                f"Error reading EEPROM. Expected the data to be of type bytes, got {type(data)}"
+            )
+        return data
+
+    def write_eeprom(
+        self, address: int, data: bytes, timeout: int = DEFAULT_EEPROM_OPERATION_TIMEOUT_uS
+    ) -> None:
+        """Write to the ESC EEPROM.
+
+        Args:
+            address: EEPROM address to be written.
+            data: Data to be written. The data must be a word (2 bytes).
+            timeout: Operation timeout (microseconds). By default, 200.000 us.
+
+        """
+        self.slave.eeprom_write(address, data, timeout)
 
     @property
     def slave(self) -> "CdefSlave":
--- a/ingenialink/ethercat/servo.py
+++ b/ingenialink/ethercat/servo.py
@@ -60,6 +60,7 @@ class EthercatServo(PDOServo):
     DEFAULT_STORE_RECOVERY_TIMEOUT = 1
 
     DEFAULT_EEPROM_OPERATION_TIMEOUT_uS = 200_000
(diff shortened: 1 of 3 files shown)


pzinemanas:
Please update CHANGELOG

ingenialink/ethercat/servo.py
```diff
@@ -320,6 +323,49 @@ def set_pdo_watchdog_time(self, timeout: float) -> None:
             self.ETHERCAT_PDO_WATCHDOG, self.SECONDS_TO_MS_CONVERSION_FACTOR * timeout
         )
 
+    def read_eeprom(
+        self,
+        address: int,
+        length: int = DEFAULT_EEPROM_READ_BYTES_LENGTH,
```
pzinemanas, line 329: Should we check that length % 2 == 0?

ingenialink/ethercat/servo.py
```diff
@@ -320,6 +323,49 @@ def set_pdo_watchdog_time(self, timeout: float) -> None:
...
+        return data
+
+    def write_eeprom(
+        self, address: int, data: bytes, timeout: int = DEFAULT_EEPROM_OPERATION_TIMEOUT_uS
+    ) -> None:
+        """Write to the ESC EEPROM.
+
+        Args:
+            address: EEPROM address to be written.
+            data: Data to be written. The data length must be a multiple of 2 bytes.
```
pzinemanas, line 359: Should we check that len(data) % 2 == 0?

ingenialink/ethercat/servo.py
MartinALL91: It does not need to be a multiple of 2. Check the [test](https://github.com/ingeniamc/ingenialink-python/pull/447/files#diff-7e3660da5a421071fe4996b2da15a790b82f68bb1036f934ca9e8022e3692f66R9).

ingenialink/ethercat/servo.py
MartinALL91: It is already checked by [PySOEM](https://github.com/bnjmnp/pysoem/blob/master/pysoem/pysoem.pyx#L821), but I could add it as well.

pzinemanas (approved):

polfeliu:
Review the issue description. Manufacturing needs a file input just like twincat. You can request a sample to them

ingenialink/ethercat/servo.py
```diff
@@ -320,6 +323,59 @@ def set_pdo_watchdog_time(self, timeout: float) -> None:
             self.ETHERCAT_PDO_WATCHDOG, self.SECONDS_TO_MS_CONVERSION_FACTOR * timeout
         )
 
+    def read_eeprom(
```
polfeliu, line 326: Users should not be using this and the name can lead to confusions.
The first thing a user will think of an eeprom of the drive is the eeprom of config.

I would call this read_esc_eeprom.

MartinALL91:
> Review the issue description. Manufacturing needs a file input just like twincat. You can request a sample to them. Maybe it can be a different function that calls the write_eeprom that you wrote

Done

ingenialink/ethercat/servo.py
```diff
@@ -320,6 +324,75 @@ def set_pdo_watchdog_time(self, timeout: float) -> None:
...
+        Returns:
+            EEPROM data. The read data.
+
+        Raises:
+            ValueError: If the length to be read has an invalid value.
+
+        """
+        if length < 1:
+            raise ValueError("The minimum length is 1 byte.")
+        data = bytes()
```
polfeliu, line 349: using a bytearray for appending will be a lot more efficient, but up to you

ingenialink/ethercat/servo.py
```diff
@@ -320,6 +324,75 @@ def set_pdo_watchdog_time(self, timeout: float) -> None:
...
+        start_address = address
```
polfeliu, line 350: Is it necessary to create a new variable? In fact start_address increases on the loop so it's name loses sense

ingenialink/ethercat/servo.py
```diff
@@ -320,6 +324,75 @@ def set_pdo_watchdog_time(self, timeout: float) -> None:
...
+        Raises:
+            ValueError: If the data has the wrong size.
+
+        """
+        if len(data) % 2 != 0:
+            raise ValueError("The data length must be a multiple of 2 bytes.")
+        start_address = address
+        while data:
+            self.slave.eeprom_write(start_address, data[:2], timeout)
+            data = data[2:]
```
polfeliu, line 377: There are ways to iterate over array 2 by 2 that are o(1) instead of this that is o(2)

MartinALL91: Essentially the same.
What do you suggest?

ingenialink/ethercat/servo.py
MartinALL91: I could change it if you propose the necessary changes.

ingenialink/ethercat/servo.py

GullinBustin (approved):

PR #461, opened 2024-07-24
INGK-573 Implement a function to do a power cycle by software if possible

MartinALL91:
### Description

Properly test the store/restore functionalities by incorporating a power cycle.

Fixes # INGK-573

### Type of change

- [x] Add a fixture to perform a power cycle.
- [x] Expand the store/restore tests.

### Tests
- [x] Run tests.

### Others

### DIFF
--- a/tests/test_servo.py
+++ b/tests/test_servo.py
@@ -316,7 +316,7 @@ def test_load_configuration_to_subnode_zero(read_config, pytestconfig, connect_t
 @pytest.mark.canopen
 @pytest.mark.ethernet
 @pytest.mark.ethercat
-def test_store_parameters(connect_to_slave, request):
+def test_store_parameters(connect_to_slave, connect_to_rack_service):
     user_over_voltage_register = "DRV_PROT_USER_OVER_VOLT"
 
     servo, net = connect_to_slave
@@ -332,7 +332,10 @@ def test_store_parameters(connect_to_slave, request):
 
     time.sleep(5)
 
-    request.getfixturevalue("perform_power_cycle")
+    client = connect_to_rack_service
+    client.exposed_turn_off_ps()
+    time.sleep(1)
+    client.exposed_turn_on_ps()
 
     wait_until_alive(servo, timeout=10)
 
@@ -342,7 +345,7 @@ def test_store_parameters(connect_to_slave, request):
 @pytest.mark.canopen
 @pytest.mark.ethernet
 @pytest.mark.ethercat
-def test_restore_parameters(connect_to_slave, request, read_config, pytestconfig):
+def test_restore_parameters(connect_to_slave, connect_to_rack_service):
     user_over_voltage_register = "DRV_PROT_USER_OVER_VOLT"
 
     servo, net = connect_to_slave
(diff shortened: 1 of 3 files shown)


tests/test_servo.py
```diff
@@ -307,30 +315,59 @@ def test_load_configuration_to_subnode_zero(read_config, pytestconfig, connect_t
...
+
+    assert servo.read(user_over_voltage_register) == new_user_over_voltage_value
 
     servo.store_parameters()
 
-    value = servo.read("DRV_STATE_STATUS")
-    assert value is not None
+    time.sleep(5)
+
+    request.getfixturevalue("perform_power_cycle")
```
polfeliu, line 335: I find this a bit confusing to read.
Why don't we request on the pytest function arguments the rack service client, and then request the power cycle with the client handle there?
Although this pattern appears to be working I think fixtures are not meant to be used like that and at least I'm aware of a potential problem: Fixtures can, in some cases, be internally cached by pytest.
https://docs.pytest.org/en/stable/how-to/fixtures.html#fixtures-can-be-requested-more-than-once-per-test-return-values-are-cached

tests/test_servo.py
```diff
@@ -307,30 +315,59 @@ def test_load_configuration_to_subnode_zero(read_config, pytestconfig, connect_t
...
 @pytest.mark.ethernet
-def test_store_parameters(connect_to_slave):
+@pytest.mark.ethercat
+def test_store_parameters(connect_to_slave, request):
+    user_over_voltage_register = "DRV_PROT_USER_OVER_VOLT"
+
     servo, net = connect_to_slave
-    assert servo is not None and net is not None
+
+    initial_user_over_voltage_value = servo.read(user_over_voltage_register)
```
polfeliu, line 324: Does connect_to_slave load the config? So if we run this test N times the register does not continuously increase on each iteration?

tests/test_servo.py
```diff
@@ -307,30 +315,59 @@ def test_load_configuration_to_subnode_zero(read_config, pytestconfig, connect_t
...
+
+    servo.write(user_over_voltage_register, new_user_over_voltage_value)
+
+    assert servo.read(user_over_voltage_register) == new_user_over_voltage_value
 
     servo.store_parameters()
 
-    value = servo.read("DRV_STATE_STATUS")
-    assert value is not None
+    time.sleep(5)
```
polfeliu, line 333: No way of knowing if the store_parameters has been completed right....?

MartinALL91: I tried it but it did not work for all the drives.

tests/test_servo.py
MartinALL91: No, it does not load the config. The initial value is restored at the end of the test, it will not change if run N times.

tests/test_servo.py
MartinALL91: I did it this way so we can reuse the same fixture on different tests. 
Yes, you're right it's cached if the fixture is requested more than once on the same test. 
I just don't see a test that will need to call this fixture more than once.

tests/test_servo.py
polfeliu: Not if it fails.
Maybe do a 
```
try:
    modify_value()
finally:
    restore_value()
```

or on the fixture just do a load config at the beginning of each test.

GullinBustin (approved):

polfeliu (approved):

PR #464, opened 2024-08-01
INGK-957 If dictionary is wrong show an error instead of crash

MartinALL91:
### Description

Add default values to the device description attributes of the Dictionary class.

Fixes # INGK-957

### Type of change
- Set the default values of the firmware version, product code, revision number, and part number to None.
- Update the save configuration method accordingly.

### Tests
- [x] Run tests.

### Documentation

### Others

### DIFF
--- a/ingenialink/servo.py
+++ b/ingenialink/servo.py
@@ -397,6 +397,11 @@ class Servo:
         if subnode is not None and (not isinstance(subnode, int) or subnode < 0):
             raise ILError("Invalid subnode")
 
+        drive_info = self.info
+        prod_code = drive_info["product_code"]
+        rev_number = drive_info["revision_number"]
+        firmware_version = drive_info["firmware_version"]
+
         tree = ET.Element("IngeniaDictionary")
         header = ET.SubElement(tree, "Header")
         version = ET.SubElement(header, "Version")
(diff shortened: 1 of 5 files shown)


ingenialink/servo.py
```diff
@@ -416,9 +415,12 @@ def save_configuration(self, config_file: str, subnode: Optional[int] = None) ->
...
             device.set("PartNumber", self.dictionary.part_number)
-        device.set("ProductCode", str(prod_code))
-        device.set("RevisionNumber", str(rev_number))
-        device.set("firmwareVersion", str(self.dictionary.firmware_version))
+        if self.dictionary.product_code is not None:
+            device.set("ProductCode", str(self.dictionary.product_code))
+        if self.dictionary.revision_number is not None:
+            device.set("RevisionNumber", str(self.dictionary.revision_number))
+        if self.dictionary.firmware_version is not None:
+            device.set("firmwareVersion", self.dictionary.firmware_version)
```
GullinBustin, line 423: I think these values should be read from the drive instead of set it as None

GullinBustin (changes requested):

ingenialink/servo.py
MartinALL91: Done

polfeliu (approved):

GullinBustin (approved):

PR #465, opened 2024-08-06
INGK-960 Unload kvaser library only in connection teardown

GullinBustin:
Fix multiple drive connections with Kvaser CANopen transceiver.

Fixes # INGK-960

### Type of change

- [x] Reload Kvaser lib only of no drives are connected

### Tests
- [x] Run tests.

#### Manual test

1. Scan slaves
2. Connect to slave A
3. Scan slaves
4. Connect to Slave B
5. Scan slaves
6. Disconnect from A and B

### Documentation

### Others

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [7.3.4]
 
 ### Fixed
-- CANopen multiple drive connection
+- CANopen multiple drive connection with Kvaser
 
 ## [7.3.3] - 2024-07-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
-# [7.3.3] - 2024-07-15
+## [7.3.4]
+
+### Fixed
+- CANopen multiple drive connection
+
+## [7.3.3] - 2024-07-15
 
 ### Added
 - Methods to read/write to the ESC EEPROM.
--- a/ingenialink/canopen/network.py
+++ b/ingenialink/canopen/network.py
@@ -246,7 +246,9 @@ class CanopenNetwork(Network):
             Containing all the detected node IDs.
 
         """
+        self._init_kvaser_lib()
         if (self.__device, self.__channel) not in self.get_available_devices():
+            self._unload_kvaser_lib()
             raise ILError(
                 f"The {self.__device.upper()} transceiver is not detected. "
                 "Make sure that it's connected and its drivers are installed."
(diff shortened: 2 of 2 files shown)


MartinALL91 (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
-# [7.3.3] - 2024-07-15
+## [7.3.4]
+
+### Fixed
+- CANopen multiple drive connection
```
MartinALL91, line 6: It only happened when using a Kvaser, right?

ingenialink/canopen/network.py
```diff
@@ -246,7 +246,9 @@ def scan_slaves(self) -> List[int]:
             Containing all the detected node IDs.
 
         """
+        self._init_kvaser_lib()
```
polfeliu, line 249: Can you do a scan while connected to another slave ?

1. Scan slaves
2. Connect to slave A
3. Scan slaves
4. Connect to Slave B
5. Disconnect from A and B

CHANGELOG.md
GullinBustin: Yes, you are right, I will add it

ingenialink/canopen/network.py
GullinBustin: Yes, you can, but if you, for example, add another transceiver, will not see changes. But this is a library problem, for this reason we must reload the lib in the scans if we can.

ingenialink/canopen/network.py
```diff
@@ -1076,10 +1082,11 @@ def _get_available_kvaser_devices() -> List[Dict[str, Any]]:
...
-    def _reload_kvaser_lib() -> None:
-        """Reload the Kvaser library to refresh the connected transceivers."""
-        canInitializeLibrary = get_canlib_function("canInitializeLibrary")
-        canUnLoadLibrary = get_canlib_function("canUnloadLibrary")
-        canUnLoadLibrary()
-        canInitializeLibrary()
+    def _reload_kvaser_lib(self) -> None:
+        """Reload the Kvaser library to refresh the connected transceivers.
+        Only when no drives are connected"""
+        if self.__device == CAN_DEVICE.KVASER.value and not self.servos:
```
polfeliu, line 1088: This function now is deceiving. I would have placed this condition on _get_available_kvaser_devices, before calling reload

polfeliu (approved):

PR #472, opened 2024-09-30
INGK-973 Improve PySOEM error description

MartinALL91:
### Description

Improve PySOEM error description

Fixes # INGK-973

### Others

### DIFF
--- a/ingenialink/ethercat/servo.py
+++ b/ingenialink/ethercat/servo.py
@@ -1,7 +1,7 @@
 import os
 import time
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, List, Optional
 
 import ingenialogger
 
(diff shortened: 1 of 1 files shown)


ingenialink/ethercat/servo.py
```diff
@@ -202,28 +202,25 @@ def _handle_sdo_exception(
              timeout period.
 
         """
-        if isinstance(
```
polfeliu, line 205: Is Sdos the only place pysoem exceptions are raised? I would imagine that the process data could yield something similar...
In general I think this formatting would belong to pysoem, that we can contribute to. If not maybe put the formatting somewhere more global.
I.E. All WkErrors should be formatted in the same way, regardless where they happen.

MartinALL91: Only SDO operations can raise exceptions.

polfeliu (approved):

GullinBustin (approved):

PR #473, opened 2024-10-02
INGK-978 Improve CoE reconnection after a disconnection

MartinALL91:
### Description

 Improve CoE reconnection after a disconnection.

Fixes # INGK-978

### Type of change

- Retry to transition the slaves to the PreOp state.
- Only consider a drive as connected when in PreOp state.

### Tests
- Connect to an EtherCAT drive (enabling the network status listener)
- Continuously read a register and:
        1. Turn off the power supply
        2. Check that the read fails.
        3. Turn on the power supply
        4. Check that the read is successful.
        5. Repeat the previous steps several times.

### Documentation

### Others

### DIFF
--- a/ingenialink/ethercat/network.py
+++ b/ingenialink/ethercat/network.py
@@ -57,11 +57,15 @@ class NetStatusListener(Thread):
             for servo in self.__network.servos:
                 slave_id = servo.slave_id
                 servo_state = self.__network._get_servo_state(slave_id)
-                slave_is_alive = servo.slave.state == pysoem.NONE_STATE
-                if servo_state == NET_STATE.CONNECTED and slave_is_alive:
+                is_servo_alive = not servo.slave.state == pysoem.NONE_STATE
+                if not is_servo_alive and servo_state == NET_STATE.CONNECTED:
                     self.__network._notify_status(slave_id, NET_DEV_EVT.REMOVED)
                     self.__network._set_servo_state(slave_id, NET_STATE.DISCONNECTED)
-                if servo_state == NET_STATE.DISCONNECTED and not slave_is_alive:
+                if (
+                    is_servo_alive
+                    and servo_state == NET_STATE.DISCONNECTED
+                    and self.__network._recover_from_disconnection()
+                ):
                     self.__network._notify_status(slave_id, NET_DEV_EVT.ADDED)
                     self.__network._set_servo_state(slave_id, NET_STATE.CONNECTED)
                 time.sleep(self.__refresh_time)
@@ -220,7 +224,6 @@ class EthercatNetwork(Network):
         servo.reset_pdo_mapping()
         self.servos.append(servo)
         self._set_servo_state(slave_id, NET_STATE.CONNECTED)
-        self.subscribe_to_status(slave_id, self._recover_from_power_cycle)
         if net_status_listener:
             self.start_status_listener()
         return servo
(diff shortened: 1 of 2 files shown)


ingenialink/ethercat/network.py
```diff
@@ -57,11 +57,15 @@ def run(self) -> None:
...
-                if servo_state == NET_STATE.CONNECTED and slave_is_alive:
+                is_servo_alive = not servo.slave.state == pysoem.NONE_STATE
+                if not is_servo_alive and servo_state == NET_STATE.CONNECTED:
                     self.__network._notify_status(slave_id, NET_DEV_EVT.REMOVED)
                     self.__network._set_servo_state(slave_id, NET_STATE.DISCONNECTED)
-                if servo_state == NET_STATE.DISCONNECTED and not slave_is_alive:
+                if (
+                    is_servo_alive
+                    and servo_state == NET_STATE.DISCONNECTED
+                    and self.__network._recover_from_disconnection()
```
polfeliu, line 67: Maybe I'm not used to lazy conditional evaluation of python and it took me a while to understand that the recover_from_disconnection was not being executed every loop

ingenialink/ethercat/network.py
```diff
@@ -516,23 +519,27 @@ def _notify_status(self, slave_id: int, status: NET_DEV_EVT) -> None:
...
-        if status == NET_DEV_EVT.ADDED and self._ecat_master.state == pysoem.INIT_STATE:
-            self.__init_nodes()
-            if not self._check_node_state(self.servos, pysoem.PREOP_STATE):
-                raise ILStateError(
-                    "The communication cannot be recovered. Not all slaves reached PreOp state"
-                )
+        if self._ecat_master.state > pysoem.INIT_STATE:
+            return True
+        self.__init_nodes()
+        all_drives_in_preop = self._check_node_state(self.servos, pysoem.PREOP_STATE)
```
polfeliu, line 534: servos are only the ones you are connected to right?
This is compatible with servos in BOOT state on the network, right?

ingenialink/ethercat/network.py
```diff
@@ -516,23 +519,27 @@ def _notify_status(self, slave_id: int, status: NET_DEV_EVT) -> None:
...
-        """Transition the slaves to Pre-Op state after a power cycle.
+    def _recover_from_disconnection(self) -> bool:
+        """Recover the CoE communication after a disconnection.
+        All the connected slaves need to transitioned to the PreOp state.
 
-        Args:
-            status: The network status.
-
-        Raises:
-            ILStateError: If not all slaves reach PreOp state.
```
polfeliu, line 526: From the function name I would expect the exception to be thrown

ingenialink/ethercat/network.py
```diff
@@ -516,23 +519,27 @@ def _notify_status(self, slave_id: int, status: NET_DEV_EVT) -> None:
...
         self._ecat_master.read_state()
-        if status == NET_DEV_EVT.ADDED and self._ecat_master.state == pysoem.INIT_STATE:
-            self.__init_nodes()
-            if not self._check_node_state(self.servos, pysoem.PREOP_STATE):
-                raise ILStateError(
-                    "The communication cannot be recovered. Not all slaves reached PreOp state"
-                )
+        if self._ecat_master.state > pysoem.INIT_STATE:
+            return True
+        self.__init_nodes()
```
polfeliu, line 533: During this process, what happens if another thread requests a SDO Read?
Let's say the display widget update process

ingenialink/ethercat/network.py
MartinALL91: We don't need to raise an exception, we need to retry if the communication cannot be recovered.

ingenialink/ethercat/network.py
MartinALL91: The drive is disconnected at this point, no SDO operations will be attempted.

ingenialink/ethercat/network.py
MartinALL91: A slave cannot be connected to the network if it cannot reach the PreOp state.

polfeliu: by connected I mean "ingenialink connected".
We should be able to have a network of two drives:
- One in app, can reach PreOp -> Can connect to it.
- One in boot, cannot reach preop -> Can coexist in the network but we cannot connect to it

ingenialink/ethercat/network.py
MartinALL91: An exception will be raised.

ingenialink/ethercat/network.py
MartinALL91: I don't think we should have a servo in the network if we cannot interact with it.  In any case, this subject is out of the scope of this issue.

polfeliu: As long as it doesn't alter the current behaviour regarding this is fine

polfeliu (approved):

PR #474, opened 2024-10-14
INGK-980 Exclude the PDO related registers from the configuration file

MartinALL91:
### Description

Exclude the RxPDO and TxPDO related registers from the configuration file of CANopen drives.

Fixes # INGK-980

### Tests
- Connect to a CANopen drive.
- Generate its configuration file.
- Check that no RxPDO and TxPDO related registers are present in the file.

### Documentation

### Others

### DIFF
--- a/ingenialink/canopen/servo.py
+++ b/ingenialink/canopen/servo.py
@@ -128,9 +128,11 @@ class CanopenServo(Servo):
         if not is_register_valid:
             return is_register_valid
         # Exclude the RxPDO and TxPDO related registers
-        if register.identifier is not None and "CIA301_COMMS" in register.identifier:
-            return not is_register_valid
-        return is_register_valid
+        if register.identifier is not None and register.identifier.startswith(
+            ("CIA301_COMMS_TPDO", "CIA301_COMMS_RPDO")
+        ):
+            return False
+        return True
 
     @staticmethod
     def _monitoring_disturbance_map_can_address(address: int, subnode: int) -> int:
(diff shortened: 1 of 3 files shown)


ingenialink/canopen/servo.py
```diff
@@ -123,6 +123,15 @@ def _change_sdo_timeout(self, value: float) -> None:
         """Changes the SDO timeout of the node."""
         self.__node.sdo.RESPONSE_TIMEOUT = value
 
+    def _is_register_valid_for_configuration_file(self, register: Register) -> bool:
+        is_register_valid = super()._is_register_valid_for_configuration_file(register)
+        if not is_register_valid:
+            return is_register_valid
+        # Exclude the RxPDO and TxPDO related registers
+        if register.identifier is not None and "CIA301_COMMS" in register.identifier:
```
polfeliu, line 131: On volcano there are other registers that start with this prefix that but are with access none at least for the non-collected registers. Example: CIA301_COMMS_COM_ERR_BEH
I don't know if for some old product or a custom this can be an issue.
I would play it safe and filter it explicitly for tpdos and rpdos,

```
if register.identifier is not None and register.startswith(('CIA301_COMMS_TPDO', 'CIA301_COMMS_RPDO'):
      return False
```
I think returning False is way more explicit than return the is_register_valid flag. I know that at this point it would always be True so if inverted... but the simpler the better

polfeliu (approved):

PR #477, opened 2024-10-22
INGK-987 Allow users to subscribe to register updates

MartinALL91:
### Description

Add methods to subscribe/unsubscribe to register value updates.

Fixes # INGK-987

### Type of change

- Add the register_update_subscribe method.
- Add the register_update_unsubscribe method.
- Remove unnecessary checks in the convert_dtype_to_bytes method.

### Tests
- [x] Run tests.

### Documentation

### Others


ingenialink/utils/_utils.py
```diff
@@ -236,10 +236,6 @@ def convert_dtype_to_bytes(data: Union[int, float, str, bytes], dtype: REG_DTYPE
         and not isinstance(data, bytes)
     ):
         raise ValueError(f"Invalid value. Expected values: {VALID_BIT_REGISTER_VALUES}, got {data}")
-    if dtype == REG_DTYPE.BYTE_ARRAY_512:
```
polfeliu, line 239: Is this change needed?

ingenialink/servo.py
```diff
@@ -1250,11 +1251,14 @@ def write(
...
-        if not isinstance(data, bytes):
-            value = convert_dtype_to_bytes(data, _reg.dtype)
+        if isinstance(data, bytes):
+            data_bytes = data
         else:
-            value = data
-        self._write_raw(_reg, value, **kwargs)
+            data_bytes = convert_dtype_to_bytes(data, _reg.dtype)
+        self._write_raw(_reg, data_bytes, **kwargs)
+        if isinstance(data, bytes):
```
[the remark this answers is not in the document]
MartinALL91: The only case where the data is passed directly as bytes is when the disturbance data is written. I don't think we need to store this in the register models.

ingenialink/servo.py
```diff
@@ -1283,8 +1287,43 @@ def read(
...
+
+        Args:
+            reg: Updated register.
+            data: Updated value.
+
+        """
+        for callback in self.__register_update_observers:
+            callback(
+                self,
+                Register(reg.dtype, reg.access, reg.identifier, subnode=reg.subnode, storage=data),
```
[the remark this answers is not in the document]
MartinALL91: The register object can hold multiple information (UID, subnode, value) in a single object.

ingenialink/utils/_utils.py
MartinALL91: Not really but it does not make sense to have it.

ingenialink/servo.py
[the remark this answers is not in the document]
GullinBustin: I don't understand, creating a new register we lose info, such as register type, enums, etc... 
In addition, should we also send it if the notification comes from a read or a write?

ingenialink/utils/_utils.py
GullinBustin: because it is already bytes? but maybe in some case, you don't know the register type but you want to convert data to bytes, in this case now it will raise an exception

ingenialink/servo.py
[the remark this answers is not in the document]
MartinALL91: I did it this way to avoid modifying the servo's register objects. But now I think there should be no problems since all registers will be reread before storing a configuration file.

GullinBustin (approved):

ingenialink/servo.py
```diff
@@ -1284,8 +1286,45 @@ def read(
...
+        """Notify a register update to the observers.
+        The updated value is stored in the register's storage attribute.
+
+        Args:
+            reg: Updated register.
+            data: Updated value.
+
+        """
+        reg.storage_valid = True
+        reg.storage = data
```
GullinBustin, line 1321: Talking with @polfeliu , we thought that the storage attribute in the register should not be used. Dictionary and Register classes should be static. For the moment we don't need a cache for the registers values, and if we need it in the future maybe the registers are not the best class to implement it.

GullinBustin (changes requested):

ingenialink/servo.py
MartinALL91: Done

tests/test_servo.py
```diff
@@ -603,3 +616,30 @@ def test_disturbance_overflow(connect_to_slave, pytestconfig):
...
+    servo.register_update_subscribe(register_update_callback.register_update_test)
+
+    previous_reg_value = servo.read(user_over_voltage_uid, subnode=1)
+    assert register_update_callback.call_count == 1
+    assert register_update_callback.servo == servo
+    assert register_update_callback.register.identifier == user_over_voltage_uid
+    assert register_update_callback.value == previous_reg_value
+
+    new_reg_value = 100
+    servo.write(user_over_voltage_uid, data=new_reg_value, subnode=1)
```
polfeliu, line 637: if you want you could also test that when a write fails no notification is sent.
reading the code I think is properly implemented

GullinBustin (approved):

PR #479, opened 2024-10-29
INGK-988 Use a dataclass to represent a dictionary error

MartinALL91:
### Description

Use a dataclass to represent a dictionary error.

Fixes # INGK-988

### Type of change

- Create the DictionaryError class.
- Remove the DictionaryErrors class.

### Tests
- Update tests.

### Documentation

### Others

### DIFF
--- a/ingenialink/dictionary.py
+++ b/ingenialink/dictionary.py
@@ -132,44 +132,6 @@ class DictionaryError:
         return iter(astuple(self))
 
 
-class DictionaryErrors:
-    """Errors for the dictionary.
-
-    Args:
-        list_xdf_errors:  List of Elements from xdf file
-    """
-
-    def __init__(self, list_xdf_errors: List[ET.Element]) -> None:
-        self._list_xdf_errors = list_xdf_errors
-        self._errors: Dict[int, DictionaryError] = {}
-
-        self.load_errors()
-
-    def load_errors(self) -> None:
-        """Load errors from dictionary."""
-        for element in self._list_xdf_errors:
-            label = element.find(DICT_LABELS_LABEL)
-            if label is None:
-                logger.warning(f"Could not load label of error {element.attrib['id']}")
-                continue
-            error_id = int(element.attrib["id"], 16)
-            error_description = label.text
-            error_type = element.attrib["error_type"].capitalize()
-            error_affected_module = element.attrib["affected_module"]
-            self._errors[error_id] = DictionaryError(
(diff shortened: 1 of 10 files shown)


ingenialink/dictionary.py
```diff
@@ -114,6 +114,24 @@ def labels(self, cat_id: str) -> Dict[str, str]:
         return self._categories[cat_id]
 
 
+@dataclass
+class DictionaryError:
+    id: str
```
polfeliu, line 119: Why don't you store the parsed int also here?

ingenialink/dictionary.py
```diff
@@ -134,15 +152,16 @@ def load_errors(self) -> None:
...
                 continue
-            self._errors[int(element.attrib["id"], 16)] = [
-                element.attrib["id"],
-                element.attrib["affected_module"],
-                element.attrib["error_type"].capitalize(),
-                label.text,
-            ]
+            error_id = int(element.attrib["id"], 16)
+            error_description = label.text
+            error_type = element.attrib["error_type"].capitalize()
```
polfeliu, line 157: Parsers should not modify the values of the things they parse.
@GullinBustin Do you know if error type being used anywhere?

ingenialink/dictionary.py
```diff
@@ -213,7 +232,7 @@ class Dictionary(ABC):
     """Number of subnodes in the dictionary."""
     categories: DictionaryCategories
     """Instance of all the categories in the dictionary."""
-    errors: DictionaryErrors
+    errors: Dict[int, "DictionaryError"]
```
polfeliu, line 235: I'm find with this simplification, but then the DictionaryErrors class serves no purpose, so the parsing method should be moved here

ingenialink/dictionary.py
MartinALL91: To keep the format as it was before. If we store an int, we will have to make changes in ML3.

ingenialink/dictionary.py
MartinALL91: The error type is used in the error widget of ML3 for the category field.

ingenialink/dictionary.py
polfeliu: I would expect the GUI to do the formatting of values so the backwards compatibility only on the __iter__ method, not the new API

polfeliu (approved):

GullinBustin (approved):

PR #486, opened 2024-11-05
INGK-997 Simplify and fix register range method

MartinALL91:
### Type of change

- Simplify method.
- Fix bug.

### Tests
- [x] Run tests.

### Others

### DIFF
--- a/tests/test_register.py
+++ b/tests/test_register.py
@@ -162,6 +162,7 @@ def test_register_set_storage():
         (REG_DTYPE.S16, (-100, None), (-100, 32767), int),
         (REG_DTYPE.U32, (None, 100), (0, 100), int),
         (REG_DTYPE.S32, (None, None), (-2147483648, 2147483647), int),
+        (REG_DTYPE.FLOAT, (None, None), (-3.4e38, 3.4e38), float),
     ],
 )
 @pytest.mark.no_connection
--- a/tests/test_register.py
+++ b/tests/test_register.py
@@ -154,30 +154,23 @@ def test_register_set_storage():
     assert register.storage == storage
 
 
+@pytest.mark.parametrize(
+    "dtype, reg_range",
+    [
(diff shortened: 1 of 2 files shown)


GullinBustin:
Add some tests to check this bug won’t appear again.

tests/test_register.py
```diff
@@ -154,30 +154,25 @@ def test_register_set_storage():
...
-    dtype = REG_DTYPE.U8
-    register = Register(dtype, access)
-    assert register.range == (dtypes_ranges[dtype]["min"], dtypes_ranges[dtype]["max"])
+def test_register_range(dtype, reg_range, reg_type):
+    register = Register(dtype, REG_ACCESS.RW, reg_range=reg_range)
+
+    assert type(register.range[0]) is reg_type
+    assert type(register.range[1]) is reg_type
+    if None in reg_range:
+        assert register.range == (dtypes_ranges[dtype]["min"], dtypes_ranges[dtype]["max"])
```
GullinBustin, line 173: Why, if exist a None in the range, all the range is ignored and the maximum and the minimum of the register type is set?

GullinBustin (changes requested):

tests/test_register.py
MartinALL91: Done

tests/test_register.py
```diff
@@ -154,30 +154,23 @@ def test_register_set_storage():
     assert register.storage == storage
 
 
+@pytest.mark.parametrize(
+    "dtype, reg_range, expected_range, reg_type",
+    [
+        (REG_DTYPE.U8, (0, 100), (0, 100), int),
+        (REG_DTYPE.FLOAT, (0.0, 1.0), (0.0, 1.0), float),
```
polfeliu, line 161: I would also test FLOAT, (None, None)

GullinBustin (approved):

PR #499, opened 2024-11-14
INGK-1003 Create a new instance when merging dictionaries

MartinALL91:
### Type of change

- Create a new dictionary instance when merging dictionaries.

### Tests
- [x] Run tests.

### Documentation

### Others

### DIFF
--- a/tests/test_dictionary.py
+++ b/tests/test_dictionary.py
@@ -132,6 +132,23 @@ def test_merge_dictionaries_image():
 
 
 @pytest.mark.no_connection
+def test_merge_dictionaries_new_instance():
+    coco_dict_path = f"{PATH_RESOURCE}comkit/com-kit.xdf"
+    moco_dict_path = f"{PATH_RESOURCE}comkit/core.xdf"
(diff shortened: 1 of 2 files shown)


tests/test_dictionary.py
```diff
@@ -131,6 +131,23 @@ def test_merge_dictionaries_image():
...
+    dict_a = EthernetDictionaryV2(coco_dict_path)
+    dict_b = EthernetDictionaryV2(moco_dict_path)
+    dict_c = dict_a + dict_b
+
+    assert id(dict_c) != id(dict_a)
+    assert id(dict_c) != id(dict_b)
+
+    dict_d = dict_b + dict_a
+
+    assert id(dict_d) != id(dict_a)
```
polfeliu, line 147: I'm not very familiar with how deepcopy works. Could we check that is has created clones of the objects we want, but not others, with the same id checking you're doing.
For example a register must be copied, An enum instance (let's say a datatype) should not...

MartinALL91: Done

GullinBustin (approved):

polfeliu (approved):

PR #510, opened 2025-01-08
INGK-1014 Review static analysis config and fix

MartinALL91:
### Description

Apply most of the available rules. 

Some rules are intentionally ignored.

Some rules are ignored until they're fixed in later issues.

Fixes # INGK-1014

### Type of change

- Update ruff.toml.
- Fix errors.

### Documentation

- [x] Use the ruff package to format the code: `ruff format ingenialink tests virtual_drive`.
- [x] Use the ruff package to lint the code: `ruff check ingenialink tests virtual_drive`.

### Others


ruff.toml
```diff
@@ -1,11 +1,50 @@
...
+ignore = [
+    "ANN", # flake8-annotations
+    "C90", # mccabe complexity
+    "PTH", # Will be fixed in INGK-1020
+    "N801", # Will be fixed in INGK-1022
+    "D100", # Missing docstring in public module
+    "D104", # Missing docstring in public package
+    "D107", # Missing docstring in __init__
+    "FA100", # Missing `from __future__ import annotations`
+    "FIX002", # TODO comment
```
polfeliu, line 21: We should remove TODO comments and link jira issues instead.

polfeliu: would also remove TD002 and TD003

ruff.toml
```diff
@@ -1,11 +1,50 @@
...
+    "COM812", # Trailing comma. Already enforced by the formatter.
+    "ISC001", # Implicitly concatenated strings. Already enforced by the formatter.
+    "S602", # Subprocess call with shell enabled.
+    "S318", # Insecure XML parser.
+    "FBT001", # Boolean positional argument
+    "FBT002", # Boolean positional argument
+]
+[lint.per-file-ignores]
+"ingenialink/network.py" = [
+    "D102", # Missing docstring in public method
```
polfeliu, line 37: Why for only this file?

ruff.toml
```diff
@@ -1,11 +1,50 @@
...
+    "PLR2004", # Magic value comparison
+    "PT004", # Missing fixture name underscore
+    "INP001", # Missing __init__.py file
 ]
-pydocstyle.convention = "google"
\ No newline at end of file
+"virtual_drive/*" = [
+    "D", # Docstring rules
+    "ERA001", # Commented-out code
+]
```
polfeliu, line 50: Virtual drive should not ignore same rules. Create issues for this if it's too much work

polfeliu:
We should also review mypy config.
follow_imports = skip,  should be removed for sure.

I also don't like these two:
disallow_subclassing_any = False
implicit_reexport = True

MartinALL91:
> We should also review mypy config. follow_imports = skip, should be removed for sure, as it was on motionlab
> 
> I also don't like these two: disallow_subclassing_any = False implicit_reexport = True

I think it would be best to do it in a separate issue, this PR has too many changes already.

ingenialink/servo.py
```diff
@@ -893,7 +893,6 @@ def state_wait_change(self, state: SERVO_STATE, timeout: int, subnode: int = 1)
             time_diff = current_time - start_time
             if time_diff > timeout:
                 raise ILTimeoutError
-            # TODO: Remove this try-except after CAP-924 is solved.
```
polfeliu, line 837: Shouldn't we have a comment here why we have the try except?
The rule is to no have "TODO". But we can keep links to future issues with possible improvements

tests/ethernet/test_ethernet_network.py
```diff
@@ -92,14 +91,12 @@ def test_load_firmware_no_connection():
 @pytest.mark.skip()
 @pytest.mark.no_connection()
 def test_load_firmware_wrong_user_pwd():
-    # TODO: implement
```
polfeliu, line 91: Create issues for the pending work of these tests

ingenialink/servo.py
MartinALL91: I don't like the idea of a link a Jira the issue here. The Jira issue has a link to the comment.

tests/ethernet/test_ethernet_network.py
MartinALL91: The issue has already been created (INGK-1025).

ingenialink/eoe/network.py
```diff
@@ -129,43 +134,44 @@ def connect_to_slave(  # type: ignore [override]
...
+        try:
+            for ip_addr, slave_id in self._configured_slaves.items():
                 self._configure_slave(slave_id, ip_addr)
-            except ILError as e:
-                logger.error(e)
+        except ILError:
+            logger.exception("Error reconfiguring drives")
 
-    def disconnect_from_slave(self, servo: EthernetServo) -> None:  # type: ignore [override]
+    def disconnect_from_slave(self, servo: EthernetServo) -> None:  # type: ignore [override] # noqa: D102
```
GullinBustin, line 158: Doesn't this function inherit the parent class's docstring?

ingenialink/ethernet/network.py
```diff
@@ -39,20 +38,21 @@ class NetStatusListener(Thread):
...
     def run(self) -> None:
+        """Check the network status."""
         while not self.__stop:
             for servo in self.__network.servos:
                 unsuccessful_pings = 0
                 servo_ip = servo.ip_address
                 servo_state = self.__network.get_servo_state(servo_ip)
                 while unsuccessful_pings < self.__max_unsuccessful_pings:
-                    response = servo.is_alive()  # TODO: Use ping after CAP-924 is fixed
+                    response = servo.is_alive()
```
GullinBustin, line 55: Why do you removed TODO comment, I known it is not the pretties way to remember an issue, but sometimes can help. What rule forbids it?

ingenialink/exceptions.py
```diff
@@ -1,74 +1,50 @@
...
-
 
 class ILConfigurationError(ILError):
     """IngeniaLink configuration error."""
 
-    pass
-
 
-class ILUDPException(ILError):
+class ILUDPExceptionError(ILError):
```
GullinBustin, line 9: Do we use it somewhere?

tests/test_dictionary.py
```diff
@@ -88,46 +92,47 @@ def test_dictionary_description_fail(dict_path, interface, raises):
...
 @pytest.mark.parametrize(
-    "dictionary_class, dictionary_path",
+    ("dictionary_class", "dictionary_path"),
     [
         (CanopenDictionaryV2, f"{PATH_RESOURCE}canopen/test_dict_can.xdf"),
         (EthernetDictionaryV2, f"{PATH_RESOURCE}ethernet/test_dict_eth.xdf"),
     ],
 )
-@pytest.mark.no_connection
+@pytest.mark.no_connection()
```
GullinBustin, line 114: I didn't knwon that can be done, i don't like it... In pytest oficial documentation fixtures are used without parenthesis

GullinBustin (changes requested):

ingenialink/ethernet/network.py
MartinALL91: [FIX002](https://docs.astral.sh/ruff/rules/line-contains-todo/#line-contains-todo-fix002).
I created issues and linked them to the code.

ingenialink/exceptions.py
MartinALL91: Yes, on the [UDP ](https://github.com/ingeniamc/ingenialink-python/blob/7.4.0/ingenialink/utils/udp.py#L88)module.

tests/test_dictionary.py
MartinALL91: [PT023](https://docs.astral.sh/ruff/rules/pytest-incorrect-mark-parentheses-style/#pytest-incorrect-mark-parentheses-style-pt023) We can undo these changes.

MartinALL91: Done

ingenialink/eoe/network.py
MartinALL91: Fixed by adding the override decorator.

ingenialink/ethernet/network.py
GullinBustin: Ok

ingenialink/exceptions.py
GullinBustin: And it should always be caught by this except, right?  https://github.com/ingeniamc/ingenialink-python/blob/662718e54dd266f43222a56324e55c126c16c6a2/ingenialink/ethernet/network.py#L181
I don't see any way for the user to catch this exception. If this exception could be seen by a user, the rename could be dangerous.

MartinALL91: I think that the renaming is safe. The ILUDPExceptionError exception is not seen by the users as the exception is re-raised as a [ILFirmwareLoadError](https://github.com/ingeniamc/ingenialink-python/blob/662718e54dd266f43222a56324e55c126c16c6a2/ingenialink/ethernet/network.py#L183) exception (which was not renamed).

GullinBustin (approved):

ingenialink/canopen/dictionary.py
```diff
@@ -17,7 +17,7 @@ class CanopenDictionaryV2(DictionaryV2):
 
     """
 
-    _MONITORING_DISTURBANCE_REGISTERS: List[CanopenRegister] = [
+    _MONITORING_DISTURBANCE_REGISTERS: list[CanopenRegister] = [  # noqa: RUF012
```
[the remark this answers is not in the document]
MartinALL91: Done, INGK-1032

ingenialink/canopen/network.py
```diff
@@ -213,22 +217,23 @@ class CanopenNetwork(Network):
     PRODUCT_CODE_SUB_IX = 2
     REVISION_NUMBER_SUB_IX = 3
     NODE_GUARDING_PERIOD_S = 1
+    TRANSCEIVER_NOT_INSTALLED_ERROR_CODE = 126
```
polfeliu, line 220: Why should this attribute be public?

MartinALL91: I just made it in accordance with the other attributes. I will make it private.

ingenialink/canopen/network.py
```diff
@@ -239,18 +244,21 @@ def __init__(
...
             Containing all the detected node IDs.
 
         """
         if (self.__device, self.__channel) not in self.get_available_devices():
-            raise ILError(
+            msg = (
                 f"The {self.__device.upper()} transceiver is not detected. "
                 "Make sure that it's connected and its drivers are installed."
             )
+            raise ILError(
```
polfeliu, line 259: I don't understand this change. Why create a message variable and then raise it with a trailing comma?

ingenialink/canopen/network.py
```diff
@@ -266,8 +274,8 @@ def scan_slaves(self) -> List[int]:
         self._connection.scanner.reset()
         try:
             self._connection.scanner.search()
-        except Exception as e:
-            logger.error("Error searching for nodes. Exception: {}".format(e))
+        except Exception:
+            logger.exception("Error searching for nodes.")
```
polfeliu, line 278: Why remove information from the exception?

MartinALL91: The exception is automatically logged.

ingenialink/canopen/network.py
MartinALL91: Check [EM102](https://docs.astral.sh/ruff/rules/f-string-in-exception/)
I think that the trailing comma is the formatter's work.

GullinBustin:
Instead of accepting all the rules, we can activate only some of them. I chose these ones. What do you think?
`F
E
W
C90
I
N
D
UP
T10
C4
FBT
A
ISC
LOG
PIE
T20
Q
RSE
SIM
TID
TC
ARG
TD
ERA
PGH
FLY
NPY
PERF
DOC
`

ingenialink/ethercat/servo.py
```diff
@@ -222,15 +223,17 @@ def _handle_sdo_exception(
             reason += error_description
         else:
             reason = str(exception)
-        raise ILIOError(f"{default_error_msg}. {reason}") from exception
+        msg = f"{default_error_msg}. {reason}"
+        raise ILIOError(msg) from exception
 
-    def _monitoring_read_data(self, **kwargs: Any) -> bytes:
+    def _monitoring_read_data(self, **kwargs: Any) -> bytes:  # noqa: ARG002
```
polfeliu, line 229: What is the true purpose of those arguments that are passed to read method?

MartinALL91:
[New PR](https://github.com/ingeniamc/ingenialink-python/pull/510)

PR #521, opened 2025-02-04
Ingk 1040 do not allow to modify pdo maps while network is in state operational

julieta-prieto:
### Pull request name 

Try to have a descriptive title. Also, set the `draft tag` in case the PR is work in progress.

### Description

Please include a summary of the change and which issue is fixed. 

Fixes # (issue)

### Type of change

- [ ] Change 1 (description)
- [ ] Change 2 (description)

### Tests
- [ ] Run tests.

### Documentation

- [ ] Use the ruff package to format the code: `ruff format ingenialink tests virtual_drive`.
- [ ] Use the ruff package to lint the code: `ruff check ingenialink tests virtual_drive`.

### Others


ingenialink/ethercat/servo.py
```diff
@@ -51,6 +52,16 @@ def __init__(self, servo: Servo, emergency_msg: "pysoem.Emergency"):
         super().__init__(servo, emergency_msg.error_code, emergency_msg.error_reg, data)
 
 
+def _servo_not_operational(func: Callable[..., None]) -> Callable[..., None]:
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):  # type: ignore[no-untyped-def]
```
polfeliu, line 57: If this is only used on one function, is it worth creating a decorator for it?

polfeliu:
Locking PDOMap object would also be a good idea. Since this object is what parses incoming data. Once operation you should not be able to add or remove from registers from it.
To implement this, we could add a reference to the servo, that is set on set_pdo_map_to_slave. This way you get access to the network to know the state of it

ingenialink/ethercat/servo.py
julieta-prieto: I am not sure if I missed some other function that might need to use this decorator as well. This is why I implemented it like this, to be able to add it to some other functions that I may have missed 😅

pytest.ini
```diff
@@ -8,3 +8,5 @@ markers =
     eoe: EtherCAT service tests
     canopen: CANopen tests
     docker: Test executed in docker
+log_cli = False
```
MartinALL91, line 11: I would not add these changes

tests/ethercat/test_pdo.py
```diff
@@ -241,6 +241,23 @@ def test_servo_add_maps(connect_to_slave, create_pdo_map):
...
+@pytest.mark.ethercat
+def test_servo_add_maps_with_wrong_pdo_state(connect_to_slave, create_pdo_map):
+    tpdo_map, rpdo_map = create_pdo_map
+    servo, _ = connect_to_slave
+
+    servo.reset_tpdo_mapping()
+    servo.reset_rpdo_mapping()
+
+    assert servo.slave.state == pysoem.PREOP_STATE
+    servo.slave.state = pysoem.OP_STATE
```
MartinALL91, line 253: In order to reach the Operational State, PDOs must be mapped and at least one cycle of process data sent. Check the [test_start_stop_pdo](https://github.com/ingeniamc/ingenialink-python/blob/7.4.1/tests/ethercat/test_pdo.py#L315) test.

pytest.ini
julieta-prieto: Why not?
It is to allow to see the logs when running the pytests, and it is `False` by default, so they won't be displayed in Jenkins.

tests/ethercat/test_pdo.py
julieta-prieto: My goal with this test was not to replicate how the PDOs can be in Operational State, but to ensure that once they are in Operational State, no modifications can be done.

Do you think that the test should be changed?

pytest.ini
MartinALL91: Because they are for debugging purposes.

GullinBustin: For me it's ok. It does not change the behavior, but if somebody wants to see the logs can show easily.

ingenialink/ethercat/servo.py
```diff
@@ -318,9 +320,11 @@ def _on_emcy(self, emergency_msg: "pysoem.Emergency") -> None:
     def set_pdo_map_to_slave(self, rpdo_maps: list[RPDOMap], tpdo_maps: list[TPDOMap]) -> None:
         for rpdo_map in rpdo_maps:
             if rpdo_map not in self._rpdo_maps:
+                rpdo_map.network_state = [self.__slave.state]
```
polfeliu, line 323: If you pass the state it will a copy of the current one
I meant passing the slave instance itself.
Something like:
````
rpdo_map.assign_to_slave(slave)
````

File: ingenialink/pdo.py [hunk 5]
```diff
@@ -202,28 +207,61 @@ class PDOMap:
...
+            raise ILPDOOperationalError(
+                "Servo is in operational state, PDOMap can not be modified."
+            )
 
     def create_item(
         self, register: Union[EthercatRegister, CanopenRegister], size_bits: Optional[int] = None
     ) -> PDOMapItem:
         """Create a new PDOMapItem.
 
+        WARNING: This operation can not be done if the servo is in operational state.
```
polfeliu, line 243: There's specific "Raises:" section for indicating exceptions:
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

ingenialink/ethercat/servo.py
julieta-prieto: I read that if you pass it as a list it will not copy it, is the workaround to passing integers as references in python, since integers are immutable.
I passed the slave first but I had issues importing CdefSlave inm `pdo.py`

ingenialink/pdo.py
```diff
@@ -404,20 +445,32 @@ def __init__(
...
         dictionary_path: str,
         servo_status_listener: bool = False,
+        network_state: list[int] = [],
     ):
         super().__init__(target, dictionary_path, servo_status_listener)
         self._rpdo_maps: list[RPDOMap] = []
         self._tpdo_maps: list[TPDOMap] = []
+        # Network state is passed by reference. Since the pysoem state is an integer,
+        # it is declared as a list of integers so that it can mutate along with the reference
+        self.__network_state: list[int] = network_state
```
polfeliu, line 455: Oh I see this now. But it's a bit of a hack :P
Having a reference to the network will solve you this problem

ingenialink/pdo.py
julieta-prieto: I am not specifying the error it raises because that is already specified in `check_servo_is_not_in_operational_state`.
I just wanted to add a warning in the docstring so that it is very clear that this should only be used if the state is not operational.

ingenialink/pdo.py
julieta-prieto: I found a way to make it work with the typehint, so I am resolving this conversation

ingenialink/pdo.py
```diff
@@ -202,28 +211,61 @@ class PDOMap:
...
+
+        Args:
+            slave: slave.
+
+        Raises:
+            ILPDOOperationalError: if the servo is in operational state.
+        """
+        if slave is None or not pysoem:
+            return
+        if slave.state == pysoem.OP_STATE:
```
MartinALL91, line 237: I think you should use the `state_check` method

julieta-prieto: I wrote it like that because I checked this: https://github.com/ingeniamc/ingenialink-python/blob/b9ef224550511fb919e80aeaaa30110354dacf01/ingenialink/ethercat/network.py#L312

So, when should we use one or the other way of checking it?

ingenialink/pdo.py
```diff
@@ -202,28 +211,61 @@ class PDOMap:
     def __init__(self) -> None:
         self.__items: list[PDOMapItem] = []
         self.__map_register_address: Optional[int] = None
+        self.__slave = None
+
+    @property
+    def slave(self) -> Optional["CdefSlave"]:
```
polfeliu, line 217: I think that the idea of this separate module was to make it generic, for ethercat and canopen idependently.
Canopen is not yet implement it, but can we try to find a way to keep it generic?

Some ideas I have not tried:
What if slave is PdoServo instead?,
can we put check_servo_is_not_in_operational_state on PdoServo as an abstract method, and then implement it on EthercatNetwork for pysoem?

ingenialink/pdo.py
MartinALL91: I would make sure to use state_check. The slave state could change and the attribute will not be updated if the state is not checked by the master.

ingenialink/ethercat/servo.py
```diff
@@ -94,13 +94,29 @@ def __init__(
...
+        """Checks if the servo is in operational state.
+
+        Raises:
+            ILPDOOperationalError: if servo is in operational state.
+        """
+        if self.slave is None or not pysoem:
+            return
+        if (
+            self.slave.state_check(pysoem.OP_STATE) == pysoem.OP_STATE
+            or self.slave.state_check(pysoem.SAFEOP_STATE) == pysoem.SAFEOP_STATE
```
polfeliu, line 114: This may be delayed due to timeout.
If the slave is in SAFEOP, the master will wait for the slave to reach OP, so the timeout will expire.
Then on the next line it will check again, although it already knows is in OP.

https://github.com/bnjmnp/SOEM/blob/9b6ebd6e43e03be461c5eab803a74786319a14db/soem/ethercatmain.c#L895

If you really don't trust the value stored in the slave structure you can refresh it, but just one time, and no waiting and timeouts needed.

````
self.slave.read_state()
if self.slave.state in [OP_STATE, SAFEOP_STAT]:
...
````

File: ingenialink/ethercat/servo.py [hunk 10]
```diff
@@ -94,13 +94,27 @@ def __init__(
...
+    def check_servo_is_not_in_operational_state(self) -> None:
+        """Checks if the servo is in operational state.
+
+        Raises:
+            ILPDOOperationalError: if servo is in operational state.
+        """
+        if self.slave is None or not pysoem:
+            return
+        self.slave.read_state()
+        if self.slave.state in [pysoem.OP_STATE, pysoem.SAFEOP_STATE]:
```
MartinALL91, line 113: Maybe we should check this the other way around (i.e. only allow to modify the PDOs in PreOp) because if the slave is in the SAFEOP_ERROR this check will fail.

ingenialink/ethercat/servo.py
julieta-prieto: @polfeliu the slave does not have `read_state` method, so the only options are to use `state_check` checking the two states, or use the suggestion that @MartinALL91 made and check if the servo is in preop state, with `state_check` again.

Otherwise, we can go back to using `self.slave.state` and check operational and safe op state.
Let me know what you prefer

polfeliu: My bad. read_state is a method of the network. If you have access it will be the same, it will refresh the state of all slaves.

But checking for PreOP seems even more correct. On the ESI file I think we have it defined with PreOp restrictions
image

ingenialink/pdo.py
```diff
@@ -202,28 +209,48 @@ class PDOMap:
     def __init__(self) -> None:
         self.__items: list[PDOMapItem] = []
         self.__map_register_address: Optional[int] = None
+        self.__slave: Optional[PDOServo] = None
+
+    @property
+    def slave(self) -> Optional["PDOServo"]:
+        """Servo state referece - allows to get the status of the network."""
```
polfeliu, line 216: This does not get the state, and it's not relevant what this property is used for.

I would state: "Servo to which this PDO is mapped."

ingenialink/pdo.py
```diff
@@ -202,28 +209,48 @@ class PDOMap:
...
             PDO Map item.
         """
+        self.__check_servo_is_in_preoperational_state()
         item = self._PDO_MAP_ITEM_CLASS(register, size_bits)
         return item
 
     def add_item(self, item: PDOMapItem) -> None:
         """Append a new item.
 
+        WARNING: This operation can not be done if the servo is in operational state.
```
polfeliu, line 248: Update all warnings: If the servo is not in pre-operational state

tests/ethercat/test_pdo.py
```diff
@@ -241,6 +241,54 @@ def test_servo_add_maps(connect_to_slave, create_pdo_map):
...
+    _, net = connect_to_slave
+
+    operation_mode_uid = "DRV_OP_CMD"
+    rpdo_registers = [operation_mode_uid]
+    operation_mode_display_uid = "DRV_OP_VALUE"
+    tpdo_registers = [operation_mode_display_uid]
+    default_operation_mode = 1
+    current_operation_mode = {}
+    new_operation_mode = {}
+    for index, servo in enumerate(net.servos):
```
polfeliu, line 255: connect_to_slave yields exactly one servo and one net. https://github.com/ingeniamc/ingenialink-python/blob/decac4aef91665785aeea7853b1efd6462026fd6/tests/conftest.py#L104

Isn't iterating through this making the test more complex?

ingenialink/pdo.py
julieta-prieto: done!

ingenialink/pdo.py
```diff
@@ -14,6 +15,12 @@
     dtype_length_bits,
 )
 
+try:
+    import pysoem
+
+except ImportError:
+    pysoem = None
```
GullinBustin, line 22: Why did you add this?

GullinBustin (changes requested):

ingenialink/pdo.py
julieta-prieto: it was needed with the previous implementation and I forgot to delete it, thanks!

ingenialink/ethercat/servo.py
```diff
@@ -94,13 +94,26 @@ def __init__(
...
         self._connection_timeout = connection_timeout
         self.__emcy_observers: list[Callable[[EmergencyMessage], None]] = []
         self.__slave.add_emergency_callback(self._on_emcy)
         super().__init__(slave_id, dictionary_path, servo_status_listener)
 
+    def check_servo_is_in_preoperational_state(self) -> None:
+        """Checks if the servo is in preoperational state.
+
+        Raises:
+            ILPDOOperationalError: if servo is not in preoperational state.
```
polfeliu, line 108: Now the exception name makes no sense.
ILPDONoPreOpError seems a tongue twister. A generic ILEcatStateError could work...

ingenialink/ethercat/servo.py
```diff
@@ -318,9 +331,11 @@ def _on_emcy(self, emergency_msg: "pysoem.Emergency") -> None:
     def set_pdo_map_to_slave(self, rpdo_maps: list[RPDOMap], tpdo_maps: list[TPDOMap]) -> None:
         for rpdo_map in rpdo_maps:
             if rpdo_map not in self._rpdo_maps:
+                rpdo_map.slave = self
```
GullinBustin, line 334: This implementation worries me. While it's currently the only way to check if the drive is operational, it could be problematic. Consider this scenario:

1. Connect an ECAT drive
2. Create some PDOs and save the PDO maps in variables
3. Disconnect the drive
4. Try to create an item in the PDO map

In step 4, the CdefSlave object likely won't exist, which could cause a crash.

GullinBustin (approved):

polfeliu (approved):

PR #534, opened 2025-02-17
Use pysoem nogil version

julieta-prieto:
### Description

GIL release control for pysoem functions.

`GilReleaseConfig` allows to configure whether GIL should always be released, or to configure the behaviour individually for:
* config init
* SDO read/write
* FOE read/write
* send/receive processdata
These arguments can still be overridden if `release_gil` argument is specified on each individual function.

### Type of change

- [X] Added GIL release control on certain EtherCAT pysoem functions.

### Tests
- [X] Run tests.

### Documentation

- [X] Update docstrings of every function, method or class that change.
- [X] Add the changes at the `[Unreleased]` section of the [CHANGELOG](CHANGELOG.md).

### Others


julieta-prieto:
closing it until pysoem PR is accepted

ingenialink/ethercat/servo.py
```diff
@@ -95,15 +97,63 @@ def __init__(
...
             raise pysoem_import_error
         self.__slave: CdefSlave = slave
         self.slave_id = slave_id
         self.__emcy_observers: list[Callable[[EmergencyMessage], None]] = []
         self.__slave.add_emergency_callback(self._on_emcy)
+        self.__slave_reference: list[CdefSlave] = []
+        self.__sdo_read_write_release_gil = sdo_read_write_release_gil
         super().__init__(slave_id, dictionary_path, servo_status_listener)
 
+    def __keep_slave_reference(self, release_gil: Optional[bool]) -> bool:
```
polfeliu, line 112: Here it's more clear that we don't need this.
Having a reference to CdefSlave and a list of CdefSlave is redundant

julieta-prieto: What do you mean with more clear @polfeliu ?

ingenialink/ethercat/network.py
```diff
@@ -131,12 +172,61 @@ def __init__(
...
+        self.__gil_release_config = gil_release_config
+        if self.__gil_release_config.always_release is not None:
+            self._ecat_master.always_release_gil = self.__gil_release_config.always_release
         timeout_us = int(1_000_000 * connection_timeout)
         self.update_sdo_timeout(timeout_us, timeout_us)
         self._ecat_master.manual_state_change = self.MANUAL_STATE_CHANGE
         self._overlapping_io_map = overlapping_io_map
         self.__is_master_running = False
         self.__last_init_nodes: list[int] = []
+        self.__ecat_master_reference: list[pysoem.CdefMaster] = []
```
polfeliu, line 184: you already have CdefMaster referenced here.
Having it redundant will also be removable by the GC, so it's not adding more protection.
If you wanted, this should be a global attribute with all the masters.

I would also try to find a mechanism to not have to add and remove it in each iteration, but I'm not sure which one that would be.
Maybe it's worth considering adding a disconnect()/delete() to the network? 
This method would check that a GIL operation is not ongoing

julieta-prieto: It is the slave that I am referencing in `servo.py`, not the master. 

How do you suggest to _"check that a GIL operation is not ongoing"_?
And why do you suggest an extra `disconnect` method?

The problem here would be if the network object is deleted from the outside. So, if we were to change the actual implementation, I would suggest to grep a reference of itself in the `__init__` method and to delete it in the `__del__` method. Though I am not 100% sure that the code will always enter `__del__` method.

polfeliu: The del method won't be helpful here. It is called when the gc has determined that is about to delete it, but I think you cannot ask him for mercy.

Here the master is "closed" https://github.com/ingeniamc/ingenialink-python/blob/9e4a10a77946c155776b8919cedad4d87276e6a6/ingenialink/ethercat/network.py#L411
Maybe it could be together with this, not sure.
I think this requires a meeting

ingenialink/ethercat/network.py
```diff
@@ -28,6 +30,69 @@
 
 logger = ingenialogger.get_logger(__name__)
 
+# Holds a reference to the Ethercat network by the time it is created -> handle no-GIL cases
+ETHERCAT_NETWORK_REFERENCES: dict[float, "EthercatNetwork"] = {}
```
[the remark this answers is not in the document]
julieta-prieto: If for some reason several networks are created, how would you know which one to remove from `ETHERCAT_NETWORK_REFERENCES` once you close the connection?

ingenialink/ethercat/network.py
```diff
@@ -28,6 +30,69 @@
...
+        creation_time_s: network creation time.
+    """
+    ETHERCAT_NETWORK_REFERENCES.pop(creation_time_s)
+
+
+@dataclass
+class GilReleaseConfig:
+    """Configuration of pysoem functions that have GIL release control."""
+
+    always_release: Optional[bool] = None
```
polfeliu, line 64: Alternatively this could be a classmethod
````
GilReleaseConfig:
...
@classmethod
def always(cls):
return cls(config_init=True, ....=True,...)


config = GilReleaseConfig.always()
````

File: ingenialink/ethercat/network.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (julieta-prieto) [reply ingenialink/ethercat/network.py:35]: @polfeliu  I modified the pytes, maybe it is clearer now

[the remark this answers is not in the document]
Comment (polfeliu) [reply ingenialink/ethercat/network.py:35]: Because each one has a specific python ID. In fact now that I see it on the test is clear that it may not work if the creation of both networks at the same time might produce conflicts. Although execution takes more than ns, python appears to not have a lot of precision: 
image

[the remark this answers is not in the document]
Comment (julieta-prieto) [reply ingenialink/ethercat/network.py]: @polfeliu  fixed

**Review by polfeliu — APPROVED** [review]

File: ingenialink/ethercat/network.py [hunk 5]
```diff
@@ -303,7 +397,7 @@ def disconnect_from_slave(self, servo: EthercatServo) -> None:  # type: ignore [
         self.servos.remove(servo)
         if not self.servos:
             self.stop_status_listener()
-            self._ecat_master.close()
+            self.close_ecat_master()
```
polfeliu, line 400: We would still be missing a no_gil lock that is waited for all gil operations are done

GullinBustin (approved):

ingenialink/ethercat/network.py
```diff
@@ -28,6 +32,72 @@
 
 logger = ingenialogger.get_logger(__name__)
 
+# Holds a reference to the Ethercat network by the time it is created -> handle no-GIL cases
```
polfeliu, line 35: "by the time it is created " no more

PR #548, opened 2025-03-11
Ingk 1068 check_node_state return False with empty servo list

lucia-teruel:
### Description

"_check_node_state" returns a True when called with an empty servos list. This sometimes leads to unhandled exceptions.

Fixes # INGK-1068

### Type of change

- [x] Function "_check_node_state" will return False if called with empty list.

### Tests
- [x] New test "test_check_node_state".
- [x] Run tests.

### Documentation

- [x] Add the changes at the `[Fixed]` section of the [CHANGELOG](CHANGELOG.md).

- [x] Use the ruff package to format the code: `ruff format ingenialink tests virtual_drive`.
- [x] Use the ruff package to lint the code: `ruff check ingenialink tests virtual_drive`.

### Others

### DIFF
--- a/ingenialink/ethercat/network.py
+++ b/ingenialink/ethercat/network.py
@@ -429,16 +429,15 @@ class EthercatNetwork(Network):
         Returns:
             True if all nodes reached the target state, else False.
         """
+        if not nodes:
+            return False
+
         node_list = nodes if isinstance(nodes, list) else [nodes]
         self._ecat_master.read_state()
 
-        return (
-            all(
-                target_state == drive.slave.state_check(target_state, ECAT_STATE_CHANGE_TIMEOUT_US)
-                for drive in node_list
-            )
-            if nodes
-            else False
+        return all(
+            target_state == drive.slave.state_check(target_state, ECAT_STATE_CHANGE_TIMEOUT_US)
+            for drive in node_list
         )
 
     def subscribe_to_status(  # type: ignore [override]
--- a/ingenialink/ethercat/network.py
+++ b/ingenialink/ethercat/network.py
@@ -667,6 +667,11 @@ class EthercatNetwork(Network):
         if self._ecat_master.state == pysoem.PREOP_STATE:
             return True
         self.__init_nodes()
+        if not self.servos:
+            log_message = (
+                "The CoE communication cannot be recovered. No slaves where detected in the network"
+            )
+            return False
         all_drives_in_preop = self._check_node_state(self.servos, pysoem.PREOP_STATE)
         if all_drives_in_preop:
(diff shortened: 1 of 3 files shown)


ingenialink/ethercat/network.py
```diff
@@ -432,9 +432,13 @@ def _check_node_state(
...
-        return all(
-            target_state == drive.slave.state_check(target_state, ECAT_STATE_CHANGE_TIMEOUT_US)
-            for drive in node_list
+        return (
+            all(
+                target_state == drive.slave.state_check(target_state, ECAT_STATE_CHANGE_TIMEOUT_US)
+                for drive in node_list
+            )
+            if nodes
+            else False
```
polfeliu, line 441: As you explain on the issue the problem is with _recover_from_disconnection function.
If you return False you will log "The CoE communication cannot be recovered. Not all slaves reached the PreOp state".

Woudn't it be better to handle this there? This special case is when servos is an empty list. 
```
def _recover_from_disconnection(self) -> bool:
       ...
        self.__init_nodes()
        if len(self.servos) == 0:
                 logger.warning("Did not detect any slave in the network")
                 return False
       all_drives_in_preop = self._check_node_state(self.servos, pysoem.PREOP_STATE)
       ...
````

File: ingenialink/ethercat/network.py [hunk 2]
```diff
@@ -432,9 +432,13 @@ def _check_node_state(
         node_list = nodes if isinstance(nodes, list) else [nodes]
         self._ecat_master.read_state()
 
-        return all(
-            target_state == drive.slave.state_check(target_state, ECAT_STATE_CHANGE_TIMEOUT_US)
-            for drive in node_list
+        return (
```
MartinALL91, line 435: I think that you should move the empty list check to the top of the function.

ingenialink/ethercat/network.py
polfeliu: I cannot make the case for one way of working of the function or the other, but python uses the other
image
My point is that this is out of context and from a maintenance point this can break

polfeliu: But functionality-wise I can make the case that if no slaves are detected the warning message:
"The CoE communication cannot be recovered. No slaves where detected in the network"
is way better than
"The CoE communication cannot be recovered. Not all slaves reached the PreOp state"
Which is misleading

lucia-teruel: @polfeliu  The `_check_node_state` function is only used on three occasions, where it only makes sense for it to be called with a non-empty list; and to return True if "all nodes have reached the target state" (which is the function's description).

polfeliu (approved):

PR #570, opened 2025-05-20
Remove subnode from ECAT/CAN devices

julieta-prieto:
### Description

Remove subnode from ECAT/CAN devices in xdf v3. 

Fixes # (issue)

### Type of change

- [ ] Use axis from CanOpenObjects instead of subnode

### Tests
- [X] Run tests.

### Documentation

### DIFF
--- a/ingenialink/dictionary.py
+++ b/ingenialink/dictionary.py
@@ -1093,8 +1093,6 @@ class DictionaryV3(Dictionary):
             if axis == 0:
                 self.subnodes[axis] = SubnodeType.COMMUNICATION
             else:
-                if axis != 1:
-                    logger.warning(f"Found {axis=}, will treat it as a Motion axis.")
                 self.subnodes[axis] = SubnodeType.MOTION
         if register.identifier is None:
             raise ValueError("Identifier must be provided.")
--- a/ingenialink/dictionary.py
+++ b/ingenialink/dictionary.py
@@ -1159,7 +1159,7 @@ class DictionaryV3(Dictionary):
         """
         object_uid = root.attrib.get(self.__UID_ATTR)
         reg_index = int(root.attrib[self.__INDEX_ATTR], 16)
-        axis = int(root.attrib[self.__AXIS_ATTR])
+        axis = int(root.attrib[self.__AXIS_ATTR]) if self.__AXIS_ATTR in root.attrib else 0
         data_type = DictionaryV3._get_canopen_object_data_type_options(
             root.attrib[self.__OBJECT_DATA_TYPE_ATTR]
         )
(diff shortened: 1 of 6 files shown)


ingenialink/dictionary.py
```diff
@@ -1085,6 +1078,28 @@ def __read_bitfields(
...
+        Raises:
+            ValueError: if register identifier is not provided.
+        """
+        if axis not in self._registers:
+            self._registers[axis] = {}
+            if axis == 0:
+                self.subnodes[axis] = SubnodeType.COMMUNICATION
+            else:
+                if axis != 1:
+                    logger.warning(f"Found {axis=}, will treat it as a Motion axis.")
```
polfeliu, line 1097: for multiaxis xdf v2 I think this must be implicit. Throwing warning is warning of something the user should not care I think...

tests/resources/canopen/test_dict_can_v3.0.xdf
```diff
@@ -19,12 +19,8 @@
 		</Categories>
 		<Devices>
 			<CANDevice firmwareVersion="2.4.1" ProductCode="61939713" PartNumber="EVS-NET-C" RevisionNumber="196617">
-				<Subnodes>
-					<Subnode index="0">Communication</Subnode>
-					<Subnode index="1">Motion</Subnode>
-				</Subnodes>
 				<CANopenObjects>
-					<CANopenObject index="0x580F" datatype="VAR" subnode="0">
+					<CANopenObject index="0x580F" datatype="VAR" axis="0">
```
GullinBustin, line 23: Are we sure about define coco as axis 0? In this case axis is equal too subnode.

julieta-prieto: I just renamed from subnode to axis. With the subnode rework, we should remove subnode and axis from ECAT/CAN devoces.

How would you suggest to define it @GullinBustin ?

polfeliu: I though axis would be an optional parameter

julieta-prieto: I think that it can not be an optional parameter right now, because we use the subnode to search for registers https://github.com/ingeniamc/ingenialink-

If we do not provide the axis, how can we find the register with this function @polfeliu  @GullinBustin ? 
python/blob/7fd167497e9b2e0fec6a10caeba88d448d4e44d2/ingenialink/dictionary.py#L447

It should be removed once the code is adapted to subnode/axis being optional

GullinBustin: For me the dictionary format and the Dictionary class functions are different things. The concept "axis 0" has no sense, then, for me, axis should be optional. In ingenialink you will add the registers with no axis as subnode 0, because now the library are implemented in this way, but in the future probably it will change.

julieta-prieto: @polfeliu  @GullinBustin fixed!

PR #571, opened 2025-05-21
INGK-1096: add do not restore registers to drive context manager

julieta-prieto:
### Description

Drive context manager was restoring all registers that were modified, even store/restore registers that need a password.

### Type of change

- [X] Ignore store/restore register changes
- [X] Add argument to specify certain registers that should be ignored in the context apart from the store/restore ones.

### Tests
- [X] Add new unit tests if it applies.
- [X] Run tests.

### Documentation

- [X] Update docstrings of every function, method or class that change.

### Others

- [X] Set fix version field in the Jira issue.


ingenialink/drive_context_manager.py
```diff
@@ -13,13 +13,29 @@ class DriveContextManager:
     """
 
     def __init__(
-        self,
-        servo: Servo,
-        axis: Optional[int] = None,
-    ):
+        self, servo: Servo, axis: Optional[int] = None, do_not_restore_registers: list[str] = []
```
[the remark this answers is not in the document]
julieta-prieto: I added it here, this was the only error

ingenialink/drive_context_manager.py
```diff
@@ -13,13 +13,29 @@ class DriveContextManager:
...
+            axis: axis to store/restore registers. If not specified, all axis will be
+            stored/restored. Defaults to None.
+            do_not_restore_registers: list of registers that should not be stored/restored.
+                Defaults to [].
+        """
         self.drive = servo
         self._axis: Optional[int] = axis
 
+        self._do_not_restore_registers: list[str] = do_not_restore_registers
+        self._do_not_restore_registers.extend([
```
polfeliu, line 36: do not extend lists that come from outside. If you call this three times with the same list these will be appended three times.

I have not found a ruff rule that checks this, but in general this should not be done

You are correctly converting this to a set. So doing a set.add()

julieta-prieto: What do you mean if you call it three times with the same external list?
That would be creating three different contexts because this only happens in class initialization. So, if I am not mistaken, in any case (set or list) the data will be added three times

GullinBustin (approved):

tests/test_drive_context_manager.py
```diff
@@ -59,3 +66,60 @@ def test_drive_context_manager_nested_contexts(setup_manager):
...
+@pytest.mark.ethernet
+@pytest.mark.ethercat
+@pytest.mark.canopen
+@pytest.mark.virtual
+def test_drive_context_manager_with_do_not_restore_registers(setup_manager):
+    servo, _, _, _ = setup_manager
+    context = DriveContextManager(servo, do_not_restore_registers=[_USER_OVER_VOLTAGE_UID])
+    assert (
+        len(context._do_not_restore_registers) == 5
+    )  # COCO-MOCO store/restore regiisters + _USER_OVER_VOLTAGE_UID
```
polfeliu, line 115: regiisters

polfeliu (approved):

PR #642, opened 2025-09-15
INGK-1184 Notify disconnection after teardown is finished

MartinALL91:
### Description

The disconnection notification is being done too early, causing the callback method to be called before INGK is finished with the connection teardown.

### Type of change

- Notify of the disconnection after the teardown is finished.

### Tests

### Documentation

### DIFF
--- a/ingenialink/canopen/network.py
+++ b/ingenialink/canopen/network.py
@@ -415,14 +415,14 @@ class CanopenNetwork(Network):
             servo: Instance of the servo connected.
 
         """
-        # Notify that disconnect_from_slave has been called
-        if servo._disconnect_callback:
-            servo._disconnect_callback(servo)
         self.stop_status_listener()
         servo.stop_status_listener()
         self.servos.remove(servo)
         if not self.servos:
             self._teardown_connection()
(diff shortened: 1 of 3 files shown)


MartinALL91:
> As far as Im aware the only usage of this is for fsoe master destruction.
> 
> Why does it matter in motionlab?
> 
> Laravel has notifications for each event action lifecycle. 
> 
> The current name of "delete" does no match neither of them, so its no obvious to me if it should go before or after.

This matters due to the delayed termination of the status listener process.

GullinBustin (approved):

polfeliu:
> > As far as Im aware the only usage of this is for fsoe master destruction.
> > Why does it matter in motionlab?
> > Laravel has notifications for each event action lifecycle. 
> > The current name of "delete" does no match neither of them, so its no obvious to me if it should go before or after.
> 
> This matters due to the delayed termination of the status listener process.

Oooh. Now I see that the disconnect callback of IM is in charge of tearing down the network. The status listener having problems I think its' just the tip of the iceberg. Makes a lot more sense now

polfeliu (approved):

PR #657, opened 2025-11-13
Ingk 1211 ethercat netstatus listener may not notice missing slaves after config init

polfeliu:
### Description

Network status listener reads the stat of all drives in the network.

The state of these drives is picked from the soem slave table, which at the same time has a slavecount that indicates how many elements of the table are valid.

When config init is done, and drives are missing, the slave count is reduced, but the reference to the slave of the table is kept although contains garbage and read state will only update the state of the slaves in the table.

Calling config init is very unusual in the middle of the application, but we have stop pdos linked to it.

And stop_pdos linked to any exception in the pdo thread.

This is explains why, when disconnecting the cable of a slave while being in OP state, sometimes we detect the disconnection, and sometimes it it not detected. When disconnecting, the pdo loop will crash for obvious reasons and do a config init, with the slave missing. The PDO thread and NetStatusListener thread are racing against each other and stop pdos might break it.

### Type of change

- [ ] Clear the slave reference one is not valid anymore
- [ ] Make ethercat NetStatusListener mark missing slave references as disconnected

### Tests

Please describe the manual tests that you ran to verify your changes if it applies. 

### Documentation

### DIFF
--- a/ingenialink/ethercat/network.py
+++ b/ingenialink/ethercat/network.py
@@ -1,5 +1,4 @@
 import atexit
-import contextlib
 import os
 import re
 import threading
@@ -159,8 +158,10 @@ class NetStatusListener(Thread):
     def run(self) -> None:
         """Check the network status continuously."""
         while not self.__stop:
-            with contextlib.suppress(Exception):
+            try:
                 self.process()
+            except Exception as e:
+                logger.exception(f"Exception occurred while processing network status: {e}")
             time.sleep(self.__refresh_time)
 
     def stop(self) -> None:


GullinBustin (approved):

ingenialink/ethercat/network.py
```diff
@@ -454,16 +464,22 @@ def __init_nodes(self, *, release_gil: Optional[bool] = None) -> None:
         nodes = self._ecat_master.config_init(release_gil=release_gil)
         self._lock.release()
         if len(self.servos):
-            self._change_nodes_state(self.servos, pysoem.PREOP_STATE)
+            self._change_nodes_state(
```
GullinBustin: I think `_change_nodes_state` and `_check_node_state` should return `False` if `servo.slave_exists` is not true. If these functions are called with a node that doesn't have a slave, they will crash. Technically, these functions return `False` if not all the slaves reach the target state, which is correct.

GullinBustin (changes requested):

ingenialink/ethercat/network.py
```diff
@@ -133,28 +134,37 @@ def __init__(self, network: "EthercatNetwork", refresh_time: float = 0.25):
...
+        """Process network status for all servos.
+
+        This method checks the status of all servos in the network and notifies
+        subscribers of any state changes (connection/disconnection).
+        """
+        self._ecat_master.read_state()
+        for servo in self.__network.servos:
+            slave_id = servo.slave_id
+            servo_state = self.__network.get_servo_state(slave_id)
+            is_servo_alive = servo.slave_exists and (servo.slave.state != pysoem.NONE_STATE)
```
polfeliu: here is an actual change

MartinALL91 (approved):

PR #658, opened 2025-11-17
INGK-1214 Update hardcoded FSoE registers for PHI: add limits, units …

lucia-teruel:
### Description

Updated FSoE hardcoded registers to match PHII ones (as much as possible).  

### Type of change

- [x] Add range to hardcoded registers (needed for PHI correct functioning).
- [x] Add description, label and/or units to hardcoded registers (to improve register explorer display).


polfeliu:
for the pdo registers there was a test that did some comparisson against a xdfv3. maybe you are interested in this test approach

polfeliu (approved):

PR #782, opened 2026-04-07
Allow save_to_xcf to save a config file without registers

gonimart:
[Doc]

Remove the register check in save_to_xcf in order to allow saving the config file without registers.

### DIFF
--- a/ingenialink/configuration_file.py
+++ b/ingenialink/configuration_file.py
@@ -585,22 +585,17 @@ class ConfigurationFile(XMLBase, ABC):
 
         Args:
             xcf_path: config file target path
-
-        Raises:
-            ValueError: the configuration has no registers
-
         """
-        if not self.registers:
-            raise ValueError("registers is empty")
         tree = ElementTree.Element(self.__ROOT_ELEMENT)
         header = ElementTree.SubElement(tree, self.__HEADER_ELEMENT)
         body = ElementTree.SubElement(tree, self.__BODY_ELEMENT)
         version = ElementTree.SubElement(header, self.__VERSION_ELEMENT)
         version.text = self.version
         device = self.device.to_xcf()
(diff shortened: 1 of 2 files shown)


ingenialink/configuration_file.py
```diff
@@ -585,22 +585,17 @@ def save_to_xcf(self, xcf_path: str) -> None:
...
         header = ElementTree.SubElement(tree, self.__HEADER_ELEMENT)
         body = ElementTree.SubElement(tree, self.__BODY_ELEMENT)
         version = ElementTree.SubElement(header, self.__VERSION_ELEMENT)
         version.text = self.version
         device = self.device.to_xcf()
-        registers_element = ElementTree.SubElement(device, self.__REGISTERS_ELEMENT)
-        for register in self.registers:
-            registers_element.append(register.to_xcf())
+        if self.registers:
+            registers_element = ElementTree.SubElement(device, self.__REGISTERS_ELEMENT)
```
polfeliu: the registers element mut continue to exist, it just needs to be empty

gonimart: agreed, I just need to remove the emptiness check and that's it, `self.registers` always exists.

polfeliu: make sure the output respects the https://github.com/ingeniamc/xcf-specification. I'm not sure if there are tests on IL repo that check that. on the specification there are

GullinBustin: It is a good point! I think Registers does not allow 0 elements. In the XSD the default value for minOccurs is 1.

gonimart: You're right. I checked and the register element can't be empty, nor the Registers container. 
image
I also tested it by adding a sample xcf without Register elements and also with/out Registers container and the schema validation fails.

So, I guess this PR shouldn't be merged, as it doesn't comply with the specification. Anyway, the test I was implementing no longer relies on this change, since it now uses the servo's methods directly.

PR #800, opened 2026-04-28
Ingk 1267 refactor test servo skip pattern to use not valid for product marker

polfeliu:
Refactors skip patterns in hardware tests to use `summit-testing-framework` product markers instead of inline `pytest.skip()` calls. This makes filtering declarative (at collection time) and consistent with the rest of the test suite.

### DIFF
--- a/tests/test_get_adapters_addresses.py
+++ b/tests/test_get_adapters_addresses.py
@@ -11,17 +11,15 @@ def adapters_module():
         import ingenialink.get_adapters_addresses as adapters  # noqa: PLC0415
 
         return adapters
-    pytest.skip(f"Skipping test, only available on Windows, platform={current_platform}")
+
+    raise NotImplementedError(
+        "Get adapters addresses is not implemented on this platform. "
+        "Do not run these tests under this platform"
+    )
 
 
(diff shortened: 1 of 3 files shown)


tests/test_get_adapters_addresses.py
```diff
@@ -11,17 +11,15 @@ def adapters_module():
         import ingenialink.get_adapters_addresses as adapters  # noqa: PLC0415
 
         return adapters
-    pytest.skip(f"Skipping test, only available on Windows, platform={current_platform}")
+
+    raise NotImplementedError(
+        "Get adapters addresses is not implemented on this platform. "
+        "Do not run these tests under this platform"
+    )
```
julieta-prieto, line 18: are you sure this is never called from Linux raising this error?

GullinBustin (approved):

MartinALL91 (approved):

tests/test_get_adapters_addresses.py
polfeliu: the purpose of the exception is to detect it if its missconfigured on ci. it passes ci so im ssure :)

PR #815, opened 2026-05-28
INGK-1258 Extract DriveRegistersState/Session from DriveContextManager

polfeliu:
## Summary

Plan: [MOT3-5567 Unified Register State Management](https://github.com/CeleraMotion/drives-rfc/blob/master/software/MOT3-5567-unified-register-state-management.md)

Extracts reusable primitives from `DriveContextManager` so MotionLab3 and other consumers can manage register state without the context-manager coupling.

## New Types

- **`DriveRegistersValue`** — immutable snapshot of register values. Can be built from hardware (`from_hardware`) or from an in-memory cache (`from_dict`). Supports `diff()` between two snapshots and `filter_registers()` for pre/post-filtering.
- **`DriveRegistersSession`** — tracks register changes against a baseline via servo callbacks. Provides `restore()` (undo tracked changes), `force_restore()` (re-read hardware + diff + restore), `current_value()` (baseline + changes), and `reset()` (clear tracked changes).
- **`RestoreResult`** — structured result with `restored`, `failed`, `skipped` lists replacing exception-based error handling.

## Refactors

- `DriveContextManager` delegates to an internal `DriveRegistersSession`; backward-compatible properties preserved.
- `_do_not_restore_registers` stores `Register` objects instead of UID strings.
- New `baseline` parameter lets callers skip the hardware read in `__enter__`.
- `Dictionary.find_registers(*patterns)` — new method accepting exact UIDs and glob patterns via `fnmatch`.
- `Register.__eq__`/`__hash__` — identity by `(subnode, identifier)`, enabling use as dict keys and set members.
- PDO exclusion uses glob `ETG_COMMS_RPDO_MAP*` / `ETG_COMMS_TPDO_MAP*` instead of substring matching, preserving ASSIGN register tracking.

## Tests

~25 new tests: `TestFindRegisters` (6), `TestRegisterEquality` (7), `TestRestoreResult` (3), `TestDriveRegistersSession` (7), integration tests for baseline/diff/state (4).
- **Integration tests** (4 tests) — baseline param, state diff, from_dict, get
- **Updated existing tests** — do_not_restore assertions use `find_registers`


ingenialink/drive_context_manager.py
```diff
@@ -26,6 +79,304 @@
...
+        return cls(register_values)
+
+    def to_tuple_dict(self) -> OrderedDict[tuple[int, str], REG_VALUE]:
+        """Convert to an OrderedDict keyed by (subnode, uid) tuples.
+
+        Returns:
+            OrderedDict mapping (subnode, uid) to register value.
+        """
+        return OrderedDict(
+            ((reg.subnode, cast("str", reg.identifier)), val) for reg, val in self._values.items()
```
julieta-prieto: If you don't use the classmethod and you directly initialize `DriveRegistersState`, you could have registers with `None` in identifiers, since I believe that nothing is preventing that (maybe I am wrong).

ingenialink/drive_context_manager.py
```diff
@@ -26,6 +79,304 @@
 _DIST_DATA_OBJECT_UID = "DIST_DATA"
 
 
+class DriveRegistersState:
```
julieta-prieto: Personally, I find `DriveRegistersValue` much more clear than `DriveRegistersState`.
You also refer to state in `diff` method when you're comparing the registers. But it is the value difference that you're checking right?

ingenialink/drive_context_manager.py
```diff
@@ -26,6 +79,304 @@
...
+                if self_value != other_value:
+                    result[register] = (self_value, other_value)
+        return result
+
+    @classmethod
+    def from_dict(cls, data: OrderedDict[Register, REG_VALUE]) -> "DriveRegistersState":
+        """Construct a state from an existing in-memory mapping.
+
+        This enables building a baseline without hardware I/O, for example
+        from MotionLab3's ``MLRegister._value`` cache.
```
julieta-prieto: It is very weird to see MotionLab3 references here... this is a public library and it may be strange for a user to find a reference to something they may not even use.

ingenialink/drive_context_manager.py
```diff
@@ -26,6 +79,304 @@
...
+        result: OrderedDict[Register, tuple[REG_VALUE, REG_VALUE]] = OrderedDict()
+        for register, self_value in self._values.items():
+            if register in other._values:
+                other_value = other._values[register]
+                if self_value != other_value:
+                    result[register] = (self_value, other_value)
+        return result
+
+    @classmethod
+    def from_dict(cls, data: OrderedDict[Register, REG_VALUE]) -> "DriveRegistersState":
```
julieta-prieto: Why doesn't this method have the option of ignoring registers?

I am not 100% sure how we'll use it. I understand that `from_hardware` method has the option of skipping them since it will read all registers.
In this case, wouldn't it be possible to have a big list of registers from which you don't care to track X,Y and Z register?

ingenialink/drive_context_manager.py
```diff
@@ -26,6 +79,304 @@
...
+        baseline: The immutable snapshot used as the reference state.
+        do_not_restore_registers: UIDs that should be ignored by the tracker.
+        axis: Axis (subnode) scope.  ``None`` means all axes.
+    """
+
+    def __init__(
+        self,
+        servo: Servo,
+        baseline: DriveRegistersState,
+        do_not_restore_registers: set[Register],
```
julieta-prieto: shouldn't this be optional?

ingenialink/drive_context_manager.py
```diff
@@ -26,6 +79,304 @@
...
+        Returns:
+            OrderedDict mapping ``(subnode, uid)`` to the changed register value.
+        """
+        return OrderedDict(
+            ((reg.subnode, cast("str", reg.identifier)), val) for reg, val in self.changes.items()
+        )
+
+    def _register_update_callback(
+        self,
+        servo: Servo,  # noqa: ARG002
```
julieta-prieto: why this noqa?

ingenialink/drive_context_manager.py
```diff
@@ -26,6 +79,304 @@
...
+        if register in self._do_not_restore_registers:
+            return
+        if register not in self.baseline._values:
+            return
+
+        if register in self.changes:
+            previous_value = self.changes[register]
+        else:
+            previous_value = self.baseline._values[register]
+        current_value = value if value is not None else previous_value
```
julieta-prieto: in which case can this callback be called with `value=None`?

ingenialink/drive_context_manager.py
```diff
@@ -26,6 +79,304 @@
...
+        if current_value == previous_value:
+            return
+        self.changes[register] = current_value
+        logger.debug(
+            f"{id(self)}: uid={register.identifier!r} changed"
+            f" from {previous_value!r} to {current_value!r}"
+        )
+
+    def state(self) -> DriveRegistersState:
+        """Return the current expected state: baseline overlaid with tracked changes.
```
julieta-prieto: what do you mean by expected state? expected by what?
isn't this the current state? Again I think that value is much more clear

ingenialink/drive_context_manager.py
```diff
@@ -53,75 +405,80 @@ def __init__(
...
+        return self._drive
+
+    @drive.setter
+    def drive(self, servo: Servo) -> None:
+        self._drive = servo
+        if self._session is not None:
+            self._session.servo = servo
+
+    @property
+    def _registers_changed(self) -> OrderedDict[tuple[int, str], REG_VALUE]:
```
julieta-prieto: if it is a property why is it protected? I don't see it used inside the class

ingenialink/drive_context_manager.py
```diff
@@ -53,75 +405,80 @@ def __init__(
...
+        """Alias for backward compatibility — delegates to the session."""
+        if self._session is None:
+            return OrderedDict()
+        return self._session.to_tuple_dict()
+
+    @property
```
julieta-prieto, line 500: if it is a property why is it protected? I don't see it used inside the class

ingenialink/drive_context_manager.py
```diff
@@ -157,45 +514,16 @@ def _complete_access_callback(
         # If at least one register is read-only, do not restore the object,
         # restore the register individually instead
         if not obj.all_registers_writable:
-            self._register_update_callback(servo=servo, register=register, value=value)
+            if self._session is None:
+                return
```
julieta-prieto: I think I'm missing something, why don't we raise an error in this case if we do so in `_register_update_callback`?

ingenialink/drive_context_manager.py
```diff
@@ -328,21 +621,16 @@ def force_restore(self, restore_registers: bool = True, restore_objects: bool =
...
 
+        assert self._session is not None
+        assert self._baseline is not None
+
         # Temporarily unsubscribe from callbacks to avoid re-populating tracking during restoration
-        self.drive.register_update_unsubscribe(self._register_update_callback)
+        self._session.stop()
         self.drive.register_update_complete_access_unsubscribe(self._complete_access_callback)
 
         try:
```
julieta-prieto, line 670: do we still need this to be inside a try-except clause?

tests/test_dictionary.py
julieta-prieto, line 1: docstrings for all tests please

julieta-prieto (commented):
Have you tried this changes with the framework, and then updating the framework in ingeniamotion for example and checking that the tests pass?

polfeliu:
> Have you tried this changes with the framework, and then updating the framework in ingeniamotion for example and checking that the tests pass?

I did previously test it but it failed catastrofically because other unrelated problems present in develop. now i see develop had 1 green build, the first in 7 days.

ingenialink/drive_context_manager.py
polfeliu: weird case but fair enough

julieta-prieto:
> > Have you tried this changes with the framework, and then updating the framework in ingeniamotion for example and checking that the tests pass?
> 
> I did previously test it but it failed catastrofically because other unrelated problems present in develop. now i see develop had 1 green build, the first in 7 days.

you can always try with ingenialink itself, tests also use the drive context manager

ingenialink/drive_context_manager.py
polfeliu: I have also detected a conflict with __hash__ of register and potential registers that are none in identifier. I'm fixing this here with narrowing but i have created the issue to make it mandatory, its a long standing issue.
the task expands to all projects
https://novantamotion.atlassian.net/browse/INGK-1271

ingenialink/drive_context_manager.py
polfeliu: required by the callback signature, but not used

ingenialink/drive_context_manager.py
polfeliu: legacy, copied from here https://github.com/ingeniamc/ingenialink-python/pull/815/changes#diff-db47e60b14e383704552df8025a25bc176bf99a8c530c3b5be1fa9ae7e166531L120

ingenialink/drive_context_manager.py
polfeliu: this is backwards compatibility. I havent found use cases other than the tests so I will try to remove it

ingenialink/drive_context_manager.py
polfeliu: unfortunately summit testing framework uses it

ingenialink/drive_context_manager.py
polfeliu: its try finally, it makes sense to me. 
let the exception be noticed but resubscribe the callbacks just in case

polfeliu:
> > > Have you tried this changes with the framework, and then updating the framework in ingeniamotion for example and checking that the tests pass?
> > 
> > I did previously test it but it failed catastrofically because other unrelated problems present in develop. now i see develop had 1 green build, the first in 7 days.
> 
> you can always try with ingenialink itself, tests also use the drive context manager

I have tested ingenilink (self), framework and testing summit

julieta-prieto (approved):

MartinALL91 (approved):

GullinBustin (approved):

PR #858, opened 2026-07-20
INGK-1284 IPv6 discovery

MartinALL91:
## Summary

Add cross-platform IPv6 device discovery using ICMPv6 multicast.

## Changes

- Use raw sockets on Linux and pcap capture on Windows.
- Handle and validate ICMPv6 Echo Replies.
- Add unit tests for discovery, interface handling, pcap capture, and packet parsing.

## Validation

Manual validation was performed on both Windows and Linux using the latest TSN firmware available:

- Verified that IPv6 discovery successfully detects connected TSN devices on Windows.
- Verified that IPv6 discovery successfully detects connected TSN devices on Linux.


ingenialink/utils/ipv6_discovery.py
```diff
@@ -0,0 +1,438 @@
...
+def discover_ipv6_devices(
+    interface: str,
+    timeout_s: float = 1.0,
+) -> list[str]:
+    r"""Discover IPv6 devices that reply to an ICMPv6 all-nodes echo request.
+
+    Args:
+        interface: Network interface to scan. On Windows, this accepts the
+            Pcap device path used by EtherCAT, for example
+            ``\\Device\\NPF_{DEADC0FF-EEEE-4444-8888-2BF6900CBFA0}``.
```
julieta-prieto: and what do we need to send in Linux?

ingenialink/utils/ipv6_discovery.py
```diff
@@ -0,0 +1,438 @@
...
+    if sys.platform != "win32":
+        raise OSError("Windows IPv6 adapters are only available on Windows.")
+
+    from ingenialink.get_adapters_addresses import (  # type: ignore[import-not-found]  # noqa: PLC0415
+        AdapterFamily,
+        ScanFlags,
+        get_adapters_addresses,
+    )
+
+    adapters = get_adapters_addresses(
```
julieta-prieto, line 193: Why are you using this method? Can't we use `pysoem.find_adapters()`? https://github.com/ingeniamc/ingenialink-python/blob/e4a79dcf61acc31126307d8351f82b359aa4410b/ingenialink/ethercat/network.py#L362

ingenialink/utils/ipv6_discovery.py
```diff
@@ -0,0 +1,438 @@
...
+    return [
+        _WindowsIpv6Adapter(
+            AdapterName=adapter.AdapterName,
+            Ipv6IfIndex=adapter.Ipv6IfIndex,
+        )
+        for adapter in adapters
+    ]
+
+
+def _is_echo_reply(response: bytes, identifier: int) -> bool:
```
julieta-prieto, line 207: can you add a docstring?

tests/test_ipv6_discovery.py
```diff
@@ -0,0 +1,261 @@
...
+
+    Returns:
+        A socket mock configured for use in a ``with`` statement.
+    """
+    socket_mock = mocker.MagicMock()
+    socket_mock.__enter__.return_value = socket_mock
+    return socket_mock
+
+
+def test_discover_ipv6_devices_collects_unique_responses(mocker, discovery_socket):
```
julieta-prieto, line 44: can you add a docstring?

tests/test_ipv6_discovery.py
```diff
@@ -0,0 +1,261 @@
...
+    assert (
+        _get_echo_reply_source_address(
+            _ethernet_ipv6_header(source_address, 58) + _echo_reply_packet(123),
+            123,
+        )
+        == source_address
+    )
+
+
+def test_pcap_capture_applies_ipv6_filter(mocker):
```
julieta-prieto: can you add a docstring?

ingenialink/utils/ipv6_discovery.py
```diff
@@ -0,0 +1,438 @@
...
+
+    with socket.socket(socket.AF_INET6, socket.SOCK_RAW, socket.IPPROTO_ICMPV6) as discovery_socket:
+        discovery_socket.setsockopt(
+            socket.IPPROTO_IPV6,
+            socket.IPV6_MULTICAST_IF,
+            interface_index,
+        )
+        if sys.platform == "win32":
+            # Windows raw sockets cannot receive multicast ICMPv6 replies.
+            with _PcapCapture(interface) as capture:
```
polfeliu: have you considered reading this table instead? https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-getipnettable2

It has all the data you need: https://learn.microsoft.com/en-us/windows/win32/api/netioapi/ns-netioapi-mib_ipnet_row2

You still have to emit the discovery packet or something that generates traffic back to us so the OS fills the table.

Using ctypes for pcap correctly (guaranteed no crashes) is not easy and I'm uneasy with this review, a cython method that instantly reads (just memory read, nothing blocks)  a table of structs should have way less risk and be more testeable

ingenialink/utils/ipv6_discovery.py
MartinALL91: pysoem.find_adapters() does not provide the IPv6 interface index required for multicast. It only returns Npcap adapter information, which socket.if_nametoindex() cannot resolve on Windows, so the current index lookup is still needed.

ingenialink/utils/ipv6_discovery.py
MartinALL91: On Linux, the native interface name, such as eth0 or enp3s0.

ingenialink/utils/ipv6_discovery.py
MartinALL91: Replaced the manual ctypes pcap bindings with cypcap, which provides a Cython-based wrapper around Npcap. This removes the unsafe native pointer handling while preserving capture of replies to the current discovery request, unlike the Windows neighbor table, which may include cached or unrelated entries.

julieta-prieto (approved):

polfeliu:
Isn't it posssible to implement e2e tests? for example, obtain the interface index of ethernet machine with rack service and:
- check that you can find it with the get adapter address and its somewhat correct (do not hardcode specifics of that machine)
- Do a discovery and check that N nodes are found.

A smoke test here is already has a lot of more coverage for this complex code.

Otherwise specify which tests have you done

MartinALL91:
> Isn't it posssible to implement e2e tests? for example, obtain the interface index of ethernet machine with rack service and:
> 
> * check that you can find it with the get adapter address and its somewhat correct (do not hardcode specifics of that machine)
> * Do a discovery and check that N nodes are found.
> 
> A smoke test here is already has a lot of more coverage for this complex code.
> 
> Otherwise specify which tests have you done

E2E tests are planned under INGK-1283 and depend on additional TSN hardware/firmware that is still under development.

I updated the PR description to document the manual validation performed.

polfeliu (approved):

PR #863, opened 2026-07-22
INGK-1281 Bootloader TFTP protocol support

MartinALL91:
## Description

Add TFTP firmware upload support for IPv6-connected drives.

- Add `TftpUploader` for uploading `.lfu` firmware files over IPv6.
- Support interface-scoped IPv6 addresses and platform-specific networking requirements.
- Include transfer validation, error handling, and retry behavior.

## Tests

- [x] Added unit tests covering successful uploads, errors, retries, address scoping, and platform-specific behavior.
- [x] Manually validated the TFTP connection, write request, and start of the data transfer with the drive bootloader.

## Documentation

- Added docstrings for the uploader and TFTP helper methods.

> [!NOTE]
> End-to-end application transfer could not yet be validated because a valid application `.lfu` file is not currently available. The TFTP session is established successfully and the transfer starts. The drive then rejects the initial data block with the `Failed to verify product signature` error, as expected because the available `.lfu` file is not valid.

### DIFF
--- a/ingenialink/utils/ipv6_tftp.py
+++ b/ingenialink/utils/ipv6_tftp.py
@@ -6,7 +6,7 @@ import sys
 import time
 from pathlib import Path
 from types import TracebackType
-from typing import Optional, Union
+from typing import NamedTuple, Optional, Union
 
 import ingenialogger
 
@@ -37,7 +37,14 @@ TFTP_RETRIES = 3
 IPV6_NEXT_HEADER_UDP = 17
 UDP_HEADER_SIZE = 8
 
-IPv6SocketAddress = tuple[str, int, int, int]
+
+class IPv6SocketAddress(NamedTuple):
+    """IPv6 socket address matching Python's ``AF_INET6`` address tuple."""
+
+    address: str
+    port: int
+    flowinfo: int
+    scopeid: int
 
 
 class TftpUploader:
@@ -86,7 +93,7 @@ class TftpUploader:
             raise ILFirmwareLoadError("The TFTP server only accepts .lfu files.")
 
         interface_index = _get_interface_index(self._interface)
-        server_address = (self._drive_address, TFTP_PORT, 0, interface_index)
+        server_address = IPv6SocketAddress(self._drive_address, TFTP_PORT, 0, interface_index)
         logger.info(f"Uploading firmware to [{self._drive_address}%{interface_index}]:{TFTP_PORT}.")
 
         try:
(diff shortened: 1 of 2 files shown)


ingenialink/utils/ipv6_tftp.py
GullinBustin, line 1: I think this file should not be in utils, and probably the IPv6 scan shouldn't be there either. I think it would be better to create a folder for IPv6 features inside the Ethernet folder or something similar (up to you), but the utils folder is too generic. I think we should remove files from there instead of creating new ones.

MartinALL91: Agreed, utils is not the ideal location. The idea is to keep these files there temporarily until the Network and Servo classes are implemented, at which point we can reorganize the IPv6 functionality appropriately.

ingenialink/utils/ipv6_tftp.py
```diff
@@ -0,0 +1,267 @@
...
+TFTP_DATA = 3
+TFTP_ACK = 4
+TFTP_ERROR = 5
+TFTP_MODE = b"octet"
+TFTP_TIMEOUT_S = 5.0
+TFTP_RETRIES = 3
+IPV6_NEXT_HEADER_UDP = 17
+UDP_HEADER_SIZE = 8
+
+IPv6SocketAddress = tuple[str, int, int, int]
```
polfeliu: any change we can use a dataclass or a namedtuple maybe? 
This roughtly corressponds to that? https://man7.org/linux/man-pages/man7/ipv6.7.html

ingenialink/utils/ipv6_tftp.py
```diff
@@ -0,0 +1,267 @@
...
+            filename: Name of the firmware file to upload.
+            capture: Windows packet capture used to receive ACK 0.
+
+        Returns:
+            IPv6 address of the server transfer endpoint.
+
+        Raises:
+            ILFirmwareLoadError: If TFTP ACK 0 is not received.
+        """
+        write_request = TftpUploader._create_write_request(filename)
```
polfeliu: staticmethods that use itself are usually class methods

polfeliu: its weird that most of this class is  static but you end up passing tftp_socket everywhere, why isnt the socket creasted on init and convert this into a contextmanager? 

```
with TFTUploader() as tftp:
 tftp.upload(file)
```

ingenialink/utils/ipv6_tftp.py
```diff
@@ -0,0 +1,267 @@
...
+    @staticmethod
+    def _upload_blocks(
+        tftp_socket: socket.socket,
+        transfer_address: IPv6SocketAddress,
+        firmware_file: Path,
+    ) -> None:
+        """Send sequential TFTP data blocks until the final block is acknowledged."""
+        block_number = 1
+        with firmware_file.open("rb") as file:
+            while True:
```
polfeliu, line 190: how fast is this transfer? should we consider progress reporting?

ingenialink/utils/ipv6_tftp.py
```diff
@@ -0,0 +1,267 @@
...
+TFTP_MODE = b"octet"
+TFTP_TIMEOUT_S = 5.0
+TFTP_RETRIES = 3
+IPV6_NEXT_HEADER_UDP = 17
+UDP_HEADER_SIZE = 8
+
+IPv6SocketAddress = tuple[str, int, int, int]
+
+
+class TftpUploader:
```
polfeliu, line 50: any kind of link to the documentation of tftp is welcome. i imageine this is a rfc

polfeliu (approved):

julieta-prieto (approved):

PR #865, opened 2026-07-23
INGK-1280 Raw SDCP serialization testing with bootloader

MartinALL91:
This pull request implements typed SDCP message encoding and frame deserialization, together with comprehensive protocol test coverage.

## SDCP message encoding

- Added typed message classes for all supported SDCP requests, successful responses, error responses, and unknown frames.
- Added self-serialization through `bytes(message)`.

## SDCP frame deserialization

- Added `SDCPDeserializer` to parse incoming frames into their corresponding typed message objects.

## Tests

- Added tests for encoding every supported SDCP request, response, and error-message type.
- Verified generated frames against the SDCP protocol examples.
- Added deserialization tests for all supported message layouts.
- Added round-trip tests using `SDCPDeserializer.deserialize(bytes(message))`.
- Added tests for malformed and incomplete frames with expected error messages.
- Added parameterized tests for invalid types and out-of-range field values.
- Added tests for unknown-frame preservation and lossless serialization.
- Added tests for protocol-oriented hexadecimal message representations.

## Validation

- Validated SDCP frame encoding and deserialization against a servo drive by reading the registers currently available in the bootloader.
- Application-level validation is not yet possible because the drive application is still under development.

### DIFF
--- a/ingenialink/utils/sdcp.py
+++ b/ingenialink/utils/sdcp.py
@@ -1,4 +1,4 @@
-"""Serialization utilities for Servo Drives Control Protocol frames."""
+"""SDCP message encoding and frame deserialization utilities."""
 
 from __future__ import annotations
 
@@ -158,8 +158,6 @@ class _SDCPCodec:
         flags: int,
         transaction_id: int,
         payload: bytes = b"",
-        *,
-        payload_name: str = "payload",
     ) -> bytes:
         """Serialize the common SDCP header and raw operation payload.
 
@@ -172,7 +170,7 @@ class _SDCPCodec:
 
         """
         if not isinstance(payload, bytes):
-            raise TypeError(f"{payload_name} must be bytes")
+            raise TypeError("payload must be bytes")
         header = b"".join((
             _SDCPCodec.serialize_uint(_SDCPFields.OPCODE, opcode),
             _SDCPCodec.serialize_uint(_SDCPFields.FLAGS, flags),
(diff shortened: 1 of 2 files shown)


ingenialink/utils/sdcp.py
polfeliu, line 1: i dont think this should go on utils

ingenialink/utils/sdcp.py
```diff
@@ -0,0 +1,833 @@
...
+        return _SDCPCodec.serialize_frame(
+            SDCPOpcode.WRITE,
+            SDCPFlag.NONE,
+            self.transaction_id,
+            payload,
+        )
+
+
+@dataclass(frozen=True, repr=False)
+class SDCPPeriodicSubscriptionRequest(_SDCPMessage):
```
polfeliu, line 290: im still unsure about this regarding the spec. We don't need to implement it though. I'm okay with having the serializer, but i wouldn't implement api for it. we just need raw read and raw write implementations at the end

polfeliu (approved):

ingenialink/utils/sdcp.py
MartinALL91: This code is being moved to a dedicated TSN package in the next PR.

ingenialink/utils/sdcp.py
MartinALL91: I agree that we don't need an API for this yet. I'd keep the message implementation since it's part of the SDCP specification, and focus on the raw read/write functionality for now

GullinBustin (approved):

PR #871, opened 2026-07-28
INGK-1290 Implement TSN servo class

MartinALL91:
## Description

This PR adds TSN servo register access using SDCP communication over UDP/IPv6.

## Changes

- Added the SDCP connection used to exchange requests and responses.
- Added TSN servo register Read and Write operations.
- Added transaction ID management.
- Added response validation and IngeniaLink error handling.
- Added big-endian register value conversion for TSN servos.

## Note

The TSN servo temporarily uses the CANopen dictionary interface until dedicated TSN dictionary support is available in INGK-1279.

## Validation

- Added unit tests for the connection and servo behavior using mocked communication.
- Validated register access locally against the registers available in the drive bootloader.

Full application-level validation is not yet possible because the TSN application firmware is still under development.


ingenialink/ethernet/tsn/sdcp/connection.py
```diff
@@ -0,0 +1,182 @@
...
+            self._ACYCLIC_PORT,
+            0,
+            interface_index,
+        )
+        self._timeout = timeout
+        self._closed = False
+
+        socket_instance: Optional[socket.socket] = None
+        try:
+            socket_instance = socket.socket(
```
gonimart, line 60: It would be really nice if we could expose timestamps (for testing purposes). Socket options allow configuring them by enabling flag bits. Python's socket module does not expose these constants, but here you have some examples: 
```
_SOF_TIMESTAMPING_TX_HARDWARE: int = 1 << 0  # timestamp at NIC TX
_SOF_TIMESTAMPING_TX_SOFTWARE: int = 1 << 1  # timestamp at kernel send path
_SOF_TIMESTAMPING_RX_HARDWARE: int = 1 << 2  # timestamp at NIC RX
_SOF_TIMESTAMPING_RX_SOFTWARE: int = 1 << 3  # timestamp at kernel softirq
_SOF_TIMESTAMPING_SOFTWARE: int = 1 << 4  # report software timestamps
_SOF_TIMESTAMPING_RAW_HARDWARE: int = 1 << 6  # report raw (unscaled) HW timestamps
```

The socket should expose whatever timestamps are available at runtime. You can check out this file https://bitbucket.org/novantamotion/dr3247-neura-eth-cyclic-test-script/src/1b3eeda799c5b3295d8b70372afeb7c98b095767/tsn_validation/protocol/hw_timestamp.py?at=DR3247DE-110-lay-groundwork-for-when-tsn-stack-arrives#lines-101 where I managed to extract the HW timestamps to calculate jitter/latency when performing cyclic communication with an ethernet servo. Of course, you will need a TSN-enabled NIC to capture the timestamps.

MartinALL91: Agreed, this would be useful for latency and jitter testing. I’ll address it in a separate issue to keep this PR focused.

julieta-prieto (approved):

polfeliu (approved):

PR #874, opened 2026-07-29
INGK-1288 Implement TSN network class

MartinALL91:
## Summary

This PR integrates SDCP node management into `EthernetNetwork`, enabling discovery, connection, disconnection, and firmware loading for SDCP-compatible drives.

## Changes

- Add IPv6 discovery and management of `SDCPNode` instances.
- Add node-based connection, disconnection, and firmware-loading workflows.
- Preserve node instances across scans and refresh their discovery information.
- Register and track the associated `SDCPServo` in the network.
- Wait for nodes to recover in application mode after firmware loading.

## Design rationale

The node-based methods are introduced in `EthernetNetwork` as an incremental step toward the planned node/servo architecture.

The existing slave-based API creates a servo directly from a network target, combining device identification and the active communication session. The new API separates these responsibilities: `SDCPNode` represents a discovered device with a stable identity, while `SDCPServo` represents an active connection to that device.

This separation allows the network to preserve the same node across connections, disconnections, operating-mode transitions, and firmware updates:

- `scan_sdcp_nodes()` discovers nodes and refreshes their information.
- `connect_to_node()` creates and registers an `SDCPServo` for a managed node.
- `disconnect_from_node()` disconnects the node and unregisters its servo.
- `load_firmware_to_node()` coordinates firmware loading, application-mode recovery, and node updates.

The existing slave-based API remains available for compatibility. Keeping both APIs in `EthernetNetwork` supports a gradual migration toward the node/servo architecture without introducing a separate SDCP-specific network abstraction.

## Automated validation

Added unit tests for node discovery, management, connection, disconnection, firmware loading, and recovery.

## Manual validation

Manually validated the workflow with a TSN-enabled drive in bootloader mode:

- Discovered and identified the drive through IPv6.
- Registered the corresponding `SDCPNode`.
- Connected to the node and created an `SDCPServo`.
- Read the registers available in bootloader mode.
- Disconnected the node through `EthernetNetwork`.
- Exercised firmware loading up to the file transfer. A complete update and subsequent recovery could not be validated because a compatible firmware file is not yet available.


ingenialink/ethernet/network.py
```diff
@@ -456,14 +475,255 @@ def _get_servo_info_for_scan(self, ip_address: str) -> SlaveInfo:
...
+        servo = node.servo
+        if servo is None or servo not in self.servos:
+            raise ValueError("The SDCP node is not connected through this network")
+
+        servo.stop_status_listener()
+        node.disconnect()
+        self._set_servo_state(servo, NetState.DISCONNECTED)
+        self._remove_servo(servo)
+
+    def load_firmware_to_node(
```
GullinBustin: It's weird to have two different functions that load firmware in two completely different ways.

MartinALL91: We could adapt the existing load_firmware() API to support SDCP nodes, but that would require extending its target-based interface with node-specific behavior. Since the long-term plan is to migrate firmware loading to the node-based architecture, I introduced load_firmware_to_node() rather than further extending the legacy API.

ingenialink/ethernet/network.py
```diff
@@ -104,7 +121,7 @@ def stop(self) -> None:
         self.__stop = True
 
 
-class EthernetNetworkBase(Generic[EthernetServoT], Network[EthernetServoT]):
+class EthernetNetworkBase(Generic[EthernetServoT], Network[Servo]):
```
polfeliu: why drop the EthernetServoT?

ingenialink/ethernet/network.py
polfeliu: I think the point of the refactor was to not introduce this on the layer that didn't want. I was expecting this function on the node instance itself. Why isn't this already the api that we would like in the end?

```
net = EthernetNetwork()
node= net.scan_nodes()
node[0].load_firmware(file)
```

polfeliu: now, i see load_firmware is already implemented on the node. In my opinion we can simply drop this

ingenialink/ethernet/network.py
```diff
@@ -456,14 +475,255 @@ def _get_servo_info_for_scan(self, ip_address: str) -> SlaveInfo:
...
+        Args:
+            node: SDCP node expected to recover.
+            timeout: Maximum time to wait for recovery.
+            poll_interval: Delay between discovery attempts.
+
+        Raises:
+            ILFirmwareLoadError: If the node does not recover
+                in application mode within the timeout.
+        """
+        deadline = time.monotonic() + timeout
```
polfeliu: could you use timeout class? https://github.com/ingeniamc/ingenialink-python/blob/c93a719764511d1b015ad68273076b06df82b6d3/ingenialink/ethercat/network.py#L718

ingenialink/ethernet/network.py
```diff
@@ -456,14 +475,255 @@ def _get_servo_info_for_scan(self, ip_address: str) -> SlaveInfo:
...
+    """
+
+    def __init__(
+        self,
+        subnet: Optional[str] = None,
+        interface: Optional[str] = None,
+    ) -> None:
+        super().__init__(subnet=subnet)
+        self.__interface = interface
+        self._sdcp_nodes: OrderedDict[NodeIdentity, SDCPNode] = OrderedDict()
```
polfeliu: what is the point of being ordered? And why keyed by node identity? the node identity already lives inside the node right?
Is finding the node by product code and serial number a typical use case worth mantaining a hash table for it?

ingenialink/ethernet/network.py
```diff
@@ -456,14 +475,255 @@ def _get_servo_info_for_scan(self, ip_address: str) -> SlaveInfo:
...
         return NetProt.ETH
 
 
 class EthernetNetwork(EthernetNetworkBase[EthernetServo]):
-    """Network for all Ethernet communications."""
+    """Network for all Ethernet communications.
+
+    Args:
+        subnet: The subnet in CIDR notation.
+        interface: The network interface used for IPv6 communication.
```
polfeliu, line 477: what about previous ipv4 coms? according to proposal 2 there's a single ethernet network that manages both types of nodes right?

ingenialink/ethernet/network.py
polfeliu: wouldnt it make much sense to key it by ip, or even mac? which is what really makes the networking routing

ingenialink/ethernet/network.py
MartinALL91: I initially put it on the network because we need to wait for the node to reboot and refresh its information (mode, revision number, and protocol version). However, since the interface and target remain unchanged, the node can handle this itself.

MartinALL91: I removed the firmware-loading API from the network. SDCPNode.load_firmware() now handles the transfer, reboot recovery, and information refresh.

ingenialink/ethernet/network.py
MartinALL91: Done

ingenialink/ethernet/network.py
MartinALL91: Done

ingenialink/ethernet/network.py
MartinALL91: Fixed by extending the managed servo type.

ingenialink/ethernet/network.py
MartinALL91: We can derive the IPv4 subnet from the network interface, but this requires additional adapter-resolution and scanning changes. I’ll implement it in a separate issue.

julieta-prieto (approved):

ingenialink/ethernet/network.py
```diff
@@ -104,7 +112,7 @@ def stop(self) -> None:
         self.__stop = True
 
 
-class EthernetNetworkBase(Generic[EthernetServoT], Network[EthernetServoT]):
+class EthernetNetworkBase(Generic[EthernetServoT], Network[Union[EthernetServoT, SDCPServo]]):
```
polfeliu, line 115: doesnt SDCPServo inherit from EthernetServoT bound? why do we need this union?

ingenialink/ethernet/network.py
```diff
@@ -335,8 +344,6 @@ def connect_to_slave(
 
         if net_status_listener:
             self.start_status_listener()
-        else:
-            self.stop_status_listener()
```
polfeliu, line 339: why?

ingenialink/ethernet/network.py
```diff
@@ -463,7 +470,156 @@ def protocol(self) -> NetProt:
...
+
+    def connect_to_node(
+        self,
+        node: SDCPNode,
+        dictionary: str,
+        servo_status_listener: bool = False,
+        net_status_listener: bool = False,
+        disconnect_callback: Optional[Callable[[Servo], None]] = None,
+        connection_timeout: float = DEFAULT_SDCP_TIMEOUT_S,
+    ) -> SDCPServo:
```
polfeliu, line 541: if its connect_to_node why does it return a servo?
disconnect_from_node input is node

ingenialink/ethernet/network.py
MartinALL91: Because net_status_listener=False should not stop a listener started by another connection.

ingenialink/ethernet/network.py
MartinALL91: In EthernetNetwork, EthernetServoT resolves to EthernetServo, so self.servos would reject SDCPServo. The union is needed to support both types.

ingenialink/ethernet/network.py
MartinALL91: connect_to_node() returns an SDCPServo because it is the communication object used to interact with the drive and access its registers. The method name refers to the node being connected and remains consistent with disconnect_from_node().

polfeliu (approved):

PR #881, opened 2026-08-04
Ingk 1272 adapt to changes in canopen v 2 4 1

julieta-prieto:
Adapted to canopen 2.4.1 changes.


ingenialink/canopen/network.py
```diff
@@ -161,15 +165,15 @@ class CanBaudrate(Enum):
...
+class CustomListener(CanopenMessageListener):
     """Custom listener for IXXAT and KVASER connection.
 
     It is used to ignore the exceptions that occur when
     the error limit is reached.
     """
 
     def __init__(self) -> None:
-        super().__init__()
+        pass
```
julieta-prieto, line 176: network access not needed; on_message_received is overridden to no-op

ingenialink/canopen/network.py
```diff
@@ -329,7 +339,7 @@ def scan_slaves(self) -> list[int]:
         if self._connection is None:
             return []
 
-        self._connection.scanner.reset()
+        self._connection.scanner.reset()  # type: ignore[no-untyped-call]
```
julieta-prieto, line 342: Added targeted no-untyped-call ignores for APIs from canopen==2.4.1, python-can, and the Kvaser bindings that are unannotated upstream

ingenialink/canopen/network.py
```diff
@@ -340,15 +350,15 @@ def scan_slaves(self) -> list[int]:
                 and self._connection.bus is not None
                 and hasattr(self._connection.bus, "reset")
             ):
-                self._connection.bus.reset()
+                getattr(self._connection.bus, "reset")()
```
polfeliu, line 353: quite dramatic this migration. what have they done so bad that we need all these dynamic accesses and ignores? Can't we do something about it?

julieta-prieto: @polfeliu  The problem is not that the API changed dramatically, it is that `canopen 2.4.1` added a `py.typed` marker while many of its methods are still completely unannotated.

* With `canopen 2.2.0`, the package was effectively treated as untyped, so calls returned `Any` and strict `mypy` did not complain.
* With `canopen 2.4.1`, `mypy` now inspects the package. 
   * There are methods that have no annotations, so strict mode reports `no-untyped-call`.

The easiest change would be:
```
self._connection.bus.reset()  # type: ignore[attr-defined]
```

Let me know if you prefer that.

polfeliu (approved):

MartinALL91 (approved):

PR #882, opened 2026-08-04
INGK-1279 Parse ETH-SDCP dictionary on ingenialink

MartinALL91:
Adds support for parsing Ethernet SDCP dictionaries in IngeniaLink.

**Changes**

- Adds Interface.SDCP and maps it to the ETH-SDCP dictionary element.
- Introduces SDCPDictionaryV3 and SDCPRegister.
- Updates SDCPServo to use the SDCP interface.
- Extends DictionaryFactory to create SDCP dictionaries.
- Adds an SDCP .xdf3 fixture and factory regression coverage.

### DIFF
--- a/tests/resources/ethernet/sdcp_example.xdf3
+++ b/tests/resources/ethernet/sdcp_example.xdf3
@@ -14,9 +14,6 @@
     </Categories>
     <Devices>
       <ETH-SDCP firmwareVersion="2.11.0" ProductCode="57745409" PartNumber="EVS-NET" RevisionNumber="196635">
-        <Subnodes>
-          <Subnode index="0">Communication</Subnode>
-        </Subnodes>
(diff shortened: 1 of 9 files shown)


GullinBustin (approved):

tests/resources/ethernet/sdcp_example.xdf3
julieta-prieto, line 1: I guess the answer is no, but isn't there a way to avoid committing this?
@polfeliu made a big effort to remove all committed dictionaries 🥲

MartinALL91: There is currently no SDCP dictionary available yet, so we need this minimal example to test the integration. We can remove it once an official dictionary is available.

polfeliu (approved):

PR #886, opened 2026-08-06
exclude monitoring registers for CoCoMoCo ethercat

polfeliu:
### Doc
Logs of all everest ethercat jobs are filled with this kind of logs for every test
image
They are very verbose and make logs impractical to work with.

Everest does not have v3 dictionaries nor monitoring on ethercat, but the ethercat dictionary is derived from the eoe dictionary that includes these registers.

This PR adds a simple rule so that they are not converted when the ethercat dictionary is created for everest products
### Test
image
Nice log of 270 lines instead of 4360


MartinALL91 (approved):

ingenialink/ethercat/dictionary.py
```diff
@@ -475,6 +476,14 @@ def __get_cia_offset(subnode: int) -> int:
         )
 
     def _read_xdf_register(self, register: ElementTree.Element) -> Optional[EthercatRegister]:
+        # EoE dictionaries include monitoring registers that CoCoMoCo does not expose through CoE.
+        if (
+            self.product_code is not None
+            and self.product_code >> 20 == COCO_MOCO_PROJECT_NUMBER
```
GullinBustin, line 482: Should you use a mask?

polfeliu: they are the uppermost bits, a bug somwhere else that received project number 0x100 + 44 should not make this fall into this branch

julieta-prieto (approved):

PR #905, opened 2026-08-20
INGK-1302 Make Jenkins test-session setup deterministic

MartinALL91:
## Summary

This change removes an intermittent Jenkins failure caused by test-session collection depending on the execution order of the parallel Windows and Linux build branches.

## What Changed

- Windows and Linux wheels continue to build in parallel.
- All test sessions are collected and resolved in a single stage after both builds complete.
- The Linux wheel is installed before importing the test setup modules and collecting tests.
- The appropriate wheel is installed before each test suite runs.
- Linux, Windows, and hardware tests run in parallel after session resolution.

## Why

Previously, Linux collected the test information while Windows resolved its session independently in a parallel branch. When Linux completed collection first, Windows reused the available runnable-test information. When Windows reached session resolution first, it triggered test collection before `ingenialink` was installed in the Windows environment, causing a `ModuleNotFoundError`.

Collecting and resolving all sessions in one stage ensures that `ingenialink` is installed before collection and removes the dependency on branch execution order. Once all sessions are resolved, the test suites can run in parallel.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -390,6 +390,7 @@ pipeline {
                     stages {
                         stage('Unstash Linux wheel') {
                             steps {
+                                sh "git clean -fdx"
                                 unstash 'publish_wheels-linux'
                             }
                         }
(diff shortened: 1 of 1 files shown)


MartinALL91:
Installing the Windows wheel before resolution would fix the immediate import error. However, the current approach is more robust because it removes the branch-order dependency entirely by collecting and resolving all sessions in one deterministic stage.

julieta-prieto (approved):

lucia-teruel (approved):

PR #913, opened 2026-08-31
INGK-1303 test: make status word timeout check deterministic

MartinALL91:
## Description

Make `test_status_word_wait_change` deterministic by limiting it to the virtual setup. The test only verifies that an unchanged status word raises `ILTimeoutError` and does not need to run against every physical setup.

On physical drives, asynchronous status-word updates, likely triggered by state changes in previous tests, can interfere with the expected timeout.

Also simplify the assertion using `pytest.raises`.

### DIFF
--- a/ingenialink/servo.py
+++ b/ingenialink/servo.py
@@ -60,7 +60,12 @@ from ingenialink.exceptions import (
 )
 from ingenialink.register import Register
 from ingenialink.table import Table
-from ingenialink.utils._utils import convert_bytes_to_dtype, convert_dtype_to_bytes, weak_lru
+from ingenialink.utils._utils import (
+    convert_bytes_to_dtype,
+    convert_dtype_to_bytes,
+    deprecated,
+    weak_lru,
+)
 from ingenialink.utils.event import create_event
 from ingenialink.utils.timeout import Timeout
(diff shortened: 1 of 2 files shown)


julieta-prieto (commented):
>  likely triggered by state changes in previous tests, can interfere with the expected timeout

What do you mean with this? This should never happen, and if it happens, we should fix it.
Have you analyzed the Wireshark logs to verify this interference?

MartinALL91:
> > likely triggered by state changes in previous tests, can interfere with the expected timeout
> 
> What do you mean with this? This should never happen, and if it happens, we should fix it. Have you analyzed the Wireshark logs to verify this interference?

There are no Wireshark logs since this failure occurred on CANopen. The Jenkins logs confirm the status-word change, but not its cause. We’ll investigate it in a separate [issue](https://novantamotion.atlassian.net/browse/INGK-1307).

julieta-prieto (approved):

polfeliu:
Apparently this function is already unused. It's not a good pattern and we replaced it by waiting the appropiate bits. I would add a deprecation marker, so it's discouraged

lucia-teruel (approved):

PR #924, opened 2026-09-14
CIT-780 Fix CANopen recovery after power-cycle

julieta-prieto:
# Background
CANopen recovery can leave the PCAN transport in an inconsistent state after a drive power cycle.

The failure was reproduced while running `ingeniamotion` target-latch and homing tests with an EVE-XCR-C (see related PR https://github.com/ingeniamc/ingeniamotion/pull/700) on PCAN_USBBUS1:

* The PCAN driver reported repeated heavy/warning bus errors while the unified supply was off.
* After power was restored, SDO requests aborted with `0x05040000` and no response.
* The CANopen notifier eventually failed with `The CAN controller was read too late`.
* Recovery could later report communication as restored because `DRV_STATE_STATUS` became readable, even though subsequent profile-position commands were acknowledged but did not produce motion.
* The network status listener and explicit framework recovery could also attempt recovery concurrently.
* Rebound `RemoteNode` instances did not retain the servo’s EMCY callback.

The target-latch command itself was correctly generating the expected control-word transition. The evidence pointed to stale or concurrently accessed CANopen transport and node state after recovery.

# Changes
* Serialize CANopen recovery attempts with a network-level lock to prevent automatic listener recovery and explicit recovery from running concurrently.
* Stop the network status listener and node guarding before resetting the CANopen connection.
* Make listener shutdown safe when recovery is initiated from the listener thread by avoiding a self-join.
* Return from the listener processing cycle after recovery so it does not continue using stale node references.
* Tear down the existing canopen.Network and create a fresh network, bus, notifier, listener collection, and node collection instead of reconnecting the failed network object in place.
* Preserve each servo node’s object dictionary when adding the replacement node.
* Restart node guarding for each rebound node.
* Reattach the servo EMCY callback when the CanopenServo.node property is rebound.
* Allow connection and node-rebinding failures to reach recover_from_disconnection() so recovery reports failure instead of continuing with partially restored state.

# Tests
Added regression tests covering:
* Transport recreation and object-dictionary preservation.
* Node-guarding shutdown and restart.
* Recovery re-entry prevention.
* EMCY callback rebinding.


ingenialink/canopen/network.py
```diff
@@ -1156,11 +1170,19 @@ def recover_from_disconnection(self, servo: Optional[Servo] = None) -> bool:
...
             True if communication is recovered, False otherwise.
         """
+        if not self.__recovery_lock.acquire(blocking=False):
+            logger.info("CANopen recovery is already in progress")
+            return False
+
+        listener_was_started = self.is_listener_started()
+        recovery_succeeded = False
         try:
+            self.stop_status_listener()
```
polfeliu: its a bit confusing that the listener calls recover from disconnection and the same function stops it (and then starts it again). I'm not sure if theres a race condition if recover from disconnection is called from another thread that is not the listener, i'd have to dig harder. A RLock is more natural to manage that in my opinion.


Recommendation from copilot that I think makes it cleaner:

```python
from threading import Lock, RLock, Thread

...

class NetStatusListener(Thread):
    ...

    def process(self, timestamps: dict[int, float]) -> dict[int, float]:
        recovery_required = False

        # Keep the existing sleep outside the lock.
        sleep(1.5)

        with self.__network._connection_lock:
            if self.__network._connection is None:
                return timestamps

            for node_id, node in list(self.__network._connection.nodes.items()):
                ...

                if is_alive:
                    ...
                elif servo_state == NetState.DISCONNECTED:
                    recovery_required = True
                    break
                else:
                    self.__network._transition_servo_state(node_id, NetDevEvt.REMOVED)

        # Preserve lock ordering: recovery_lock, then connection_lock.
        if recovery_required:
            self.__network.recover_from_disconnection()
            return {}

        return timestamps

...

class CanopenNetwork(CanopenNetworkBase[CanopenServo]):
    ...

    def __init__(self, ...) -> None:
        ...
        self.__recovery_lock = Lock()
        self._connection_lock = RLock()

    ...

    def recover_from_disconnection(
        self, servo: Optional[Servo] = None
    ) -> bool:
        if not self.__recovery_lock.acquire(blocking=False):
            logger.info("CANopen recovery is already in progress")
            return False

        try:
            with self._connection_lock:
                self._reset_connection()

                for _ in range(self.MAX_NUMBER_SERVO_ALIVE_ATTEMPTS):
                    if all(servo.is_alive() for servo in self.servos):
                        logger.info("CANopen communication recovered.")
                        return True

                    sleep(0.1)

                logger.warning(
                    "CANopen communication recovered, "
                    "but some servos are still disconnected."
                )
                return False
        except Exception as exc:
            logger.warning(
                "Failed to recover CANopen communication: %s",
                exc,
            )
            return False
        finally:
            self.__recovery_lock.release()

    ...

    def stop_status_listener(self) -> None:
        listener = self.__listener_net_status
        if listener is None:
            return

        listener.stop()
        listener.join()
        self.__listener_net_status = None

    ...
```

julieta-prieto: I applied the listener/recovery part of the suggestion: recovery now runs after releasing the connection lock, transport replacement is serialized with a recovery-lock -> connection-lock order, heartbeat timestamps are reset after transport replacement, and the listener is no longer stopped/restarted by its own recovery path. I also kept the self-join guard and added regression tests.

I intentionally did not add a global lock around every SDO/LSS/firmware/scan operation or acquire all servo I/O locks. That is a broader synchronization change with additional deadlock and concurrency tradeoffs, so I kept this fix focused on the listener/recovery race.

polfeliu (approved):

MartinALL91 (approved):

PR #928, opened 2026-09-18
Cit 855 move get register method to ingk

julieta-prieto:
# Description
Add some utilities to `configuration_file.py` to make overriding and cloning easier.

# Changes
## `ConfigRegister` Changes
* Added `clone` method: it accepts overrides for all its attributes.
* Added `clone_with_storage`: when configuration file is loaded to the servo, if there is any value in `data` field, it is preferred over storage. See https://github.com/ingeniamc/ingenialink-python/blob/fd5b683d5d973af255f0940af627b9dde4a0128d/ingenialink/servo.py#L698
  This methods overrides `storage` and sets `data` to `None` to make sure that the intended `storage` value is used.
* Added `effective_value` cached property:
  * If `data` is specified, it converts the bytes to the register type and returns it.
  * Otherwise, it returns storage values.

## `ConfigurationFile` Changes
* Added `clone` method to `ConfigurationFile`: no override accepted.
* Added `find_registers`: finds all registers with matching uid (same that `dictionary.get_registers()`)
* Added `find_register`: finds a register with matching uid and subnode (same that `dictionary.get_register()`)
  * `weak_lru` is not used here because values can change by using override methods
  * If subnode is not provided and the register is found in several subnodes, an error is raised.

# Tests
* Added unit tests for new methods.
* Validated with https://github.com/CeleraMotion/summit-testing-framework/pull/139.


ingenialink/configuration_file.py
```diff
@@ -92,6 +97,13 @@ def _get_attribute_from_element(
 }
 
 
+class _CloneUnset:
+    """Sentinel for omitted clone arguments."""
+
+
+_CLONE_UNSET = _CloneUnset()
```
polfeliu, line 103: now that 3.15 has https://peps.python.org/pep-0661/, this is ugly, but it is what it is :)

ingenialink/configuration_file.py
```diff
@@ -748,3 +816,51 @@ def override_values(self, other: "ConfigurationFile") -> None:
                     f"Table {new_table.uid!r} (subnode {new_table.subnode}) not in target; adding."
                 )
                 self.add_config_table(new_table)
+
+    def clone(self) -> "ConfigurationFile":
+        """Returns a deep copy of this ConfigurationFile."""
+        return deepcopy(self)
```
polfeliu: deep copy can have unexpected results some times duplicating objects that you didnt expect. i think this is simple enought to hand-roll it, if not at least validate on tests that you are creating clones of the intended objects, and not duplicating things like enum keys

julieta-prieto: @polfeliu changed, see [ec4fb96](https://github.com/ingeniamc/ingenialink-python/pull/928/commits/ec4fb96ccc63e5f3195e247d16aba4a80b141a47)

GullinBustin (approved):

MartinALL91 (approved):

